PR #4, opened 2021-01-12
MSEARCH-14 Implement CQL query parser

pfilippovEPAM:
Implementation of CQL query parser with required endpoint, service, converter and repository to perform search operation to the Elasticsearch server.
Currently not all features supported that are already done in existing inventory service, however they will be implemented in the following stories.


src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
```diff
@@ -0,0 +1,146 @@
...
+        return boolQuery
+          .should(convertToQuery(node.getLeftOperand()))
+          .should(convertToQuery(node.getRightOperand()));
+      case AND:
+        return boolQuery
+          .must(convertToQuery(node.getLeftOperand()))
+          .must(convertToQuery(node.getRightOperand()));
+      case NOT:
+        return boolQuery
+          .must(convertToQuery(node.getLeftOperand()))
```
MikhailFokanov, line 132: I cannot get how does it work... Did you test it with elastic?

MikhailFokanov (approved):
Looks good.

bohdan-suprun (approved):

src/main/java/org/folio/search/model/rest/response/PagedResult.java
```diff
@@ -1,18 +1,10 @@
...
 
-  private long total;
-  private long totalPages;
+  long getTotalRecords();
 
-  private int pageSize;
-  private int pageNumber;
-
-  private List<T> content;
+  List<T> getInstances();
```
hjiebsco, line 9: Assuming the result could be instances, holdings, or items, maybe consider renaming to `getRecords` to be more generic?

src/main/java/org/folio/search/model/rest/response/SearchResult.java
```diff
@@ -1,11 +1,14 @@
...
 
 @Data
-@NoArgsConstructor
-@EqualsAndHashCode(callSuper = true)
-public class SearchResult extends PagedResult<Object> {
+@AllArgsConstructor(staticName = "of")
+public class SearchResult implements PagedResult<Map<String, Object>> {
+
+  private final long totalRecords;
+  private final List<Map<String, Object>> instances;
```
hjiebsco, line 13: Consider renaming to `records` if the same generic idea applies.

src/main/java/org/folio/search/repository/SearchRepository.java
```diff
@@ -1,24 +1,62 @@
...
+import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.folio.search.model.rest.response.SearchResult;
+import org.folio.search.model.service.CqlSearchRequest;
 import org.springframework.stereotype.Repository;
 
 /**
  * Search resource repository with set of operation to perform search operations.
  */
 @Repository
+@RequiredArgsConstructor
```
hjiebsco, line 26: Consider to do the same for `IndexRepository.java`

hjiebsco (approved):
Looks good. Left a few minor comments. Thanks.

src/main/java/org/folio/search/model/rest/response/SearchResult.java
pfilippovEPAM: In current implementation - I am trying to support the contract, that is used in existing inventory search and it returns only instances
Also, during working on swagger documentation - map has been replaced with instance object

src/main/java/org/folio/search/repository/SearchRepository.java
pfilippovEPAM: fixed

pom.xml
```diff
@@ -111,6 +111,22 @@
       <version>2.4.0</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.folio</groupId>
+      <artifactId>cql2pgjson</artifactId>
+      <version>29.5.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
pfilippovEPAM: changed to the latest version

pom.xml
```diff
@@ -264,6 +280,11 @@
       <name>FOLIO Maven repository</name>
       <url>https://repository.folio.org/repository/maven-folio</url>
     </repository>
+    <repository>
+      <id>indexdata</id>
+      <name>Index Data</name>
+      <url>http://maven.indexdata.com/</url>
+    </repository>
```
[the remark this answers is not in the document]
pfilippovEPAM: Yes, thank you, it's not used with new version

PR #5, opened 2021-01-15
MSEARCH-38: Update module descriptor.

bohdan-suprun:
https://issues.folio.org/browse/MSEARCH-38 - Update module-descriptor with new APIs

## Changes:
* Update module descriptor and add new APIs:
  * `indexes` contains all operations for indexes (create/update index, index resources)
  *  `search` contains instance search API;
* Assign permissions for each endpoint;
* Requires `instance-storage:7.5` API (it is not directly queried but needed to prevent breaking changes being applied to the instance schema without the team notification).
* Inject kafka host and port and elasticsearch host and port from env variables.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -11,7 +11,7 @@
             "POST"
           ],
           "pathPattern": "/search/index/indices",
-          "permissionsRequired": ["search.index.indexes.item.post"]
+          "permissionsRequired": ["search.index.indices.item.post"]
         },
         {
           "methods": [
@@ -51,9 +51,9 @@
   ],
   "permissionSets" : [
     {
-      "permissionName": "search.index.indexes.item.post",
-      "displayName": "Search - create an index",
-      "description": "Create an index"
+      "permissionName": "search.index.indices.item.post",
+      "displayName": "Search - creates an index",
+      "description": "Creates an index"
     },
     {
       "permissionName": "search.index.mappings.item.post",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1,29 +1,76 @@
 {
   "id": "${artifactId}-${version}",
-  "name": "The blueprint module that could be used as a template for folio spring based backend modules.",
+  "name": "Search Module",
   "provides": [
     {
-      "id": "spring-template",
-      "version": "1.0",
-      "handlers": []
-    },
-    {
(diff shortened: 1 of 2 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -1,29 +1,76 @@
...
-      "handlers": []
+      "id": "indexes",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/search/index/indices",
+          "permissionsRequired": ["search.index.indexes.item.post"]
```
hjiebsco, line 14: Should we make it consistent, `indices` vs `indexes`?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1,29 +1,76 @@
...
+    {
+      "id": "instance-storage",
+      "version": "7.5"
+    }
+  ],
+  "permissionSets" : [
+    {
+      "permissionName": "search.index.indexes.item.post",
+      "displayName": "Search - create an index",
+      "description": "Create an index"
```
hjiebsco, line 56: Either we change `create` to `creates`, or change Line 60-61, Line 65-66 and 70-71.

hjiebsco (approved):
Left minor wording change requests.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Changed to `creates`, thanks!

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Yes, agree, renamed to `indices`.

src/main/resources/application.yml
```diff
@@ -8,9 +8,9 @@ spring:
       accept-single-value-as-array: true
   elasticsearch:
     rest:
-      uris: http://elasticsearch:9200
+      uris: http://${ELASTICSEARCH_HOST:elasticsearch}:${ELASTICSEARCH_PORT:9200}
```
pfilippovEPAM, line 11: maybe better to use elasticsearch_url parameter? in case we would have external elasticsearch url with https protocol

bohdan-suprun: Make sense, but I've agreed the ENV variables names with devops and they already added them to the helm charts. Let's postpone the change until we have a case, then we can change it to a single property?

pfilippovEPAM (approved):

PR #7, opened 2021-01-20
MSEARCH-16 Implement search and sort by title

pfilippovEPAM:


src/main/java/org/folio/search/model/types/InventorySearchType.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.util.Optional;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
+
+@Getter
+@RequiredArgsConstructor
+public enum InventorySearchType {
```
MikhailFokanov, line 11: Why do you need this hardcode for TITLE? As far as I know, UI will provide the names of the fields in CQL.

MikhailFokanov: You just need to map the field name (e.g. "title"), to specific ES field(s) (e.g. "title.eng and title.original" or "title.*") having in mind what is the mapping for this field.

pfilippovEPAM: it the group of fields, From my viewpoint - it is the best for now to support existing search features in inventory

src/main/java/org/folio/search/model/metadata/PlainFieldDescription.java
```diff
@@ -38,16 +43,16 @@
    */
   private boolean languageSource;
 
+  /**
+   * Specifies if fields should be returned as part of elasticsearch response or not.
+   */
+  private boolean showInResponse;
```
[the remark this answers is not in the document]
pfilippovEPAM: It will be used after, I will implement it in the MSEARCH-18

src/main/java/org/folio/search/service/converter/ElasticsearchHitConverter.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @SuppressWarnings("unchecked")
+  private static boolean isMultiLanguageField(Object fieldValue) {
+    return fieldValue instanceof Map && ((Map<String, Object>) fieldValue).containsKey("src");
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Object getSourceMultilangValue(Object fieldValue) {
+    return (fieldValue instanceof Map) ? MapUtils.getObject((Map<String, Object>) fieldValue, "src") : null;
+  }
```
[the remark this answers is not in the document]
pfilippovEPAM: It won't take a lot of time, also, we will return from elasticsearch only five fields that will be shown, so it won't affect performance

hjiebsco (approved):
Looks fine based on my cursory check. Please also ask peer backend developers to take a look since they are more familiar with the code. Thanks.

hjiebsco:
BTW, when I run `mvn install`, sometimes it fails with some integration tests like below. You might want to give it a try.
```
[ERROR] Failures: 
[ERROR]   SearchInstanceIT.canSearchByInstanceId_wildcard:31 JSON path "totalRecords"
Expected: is <1>
     but: was <0>
[ERROR]   SearchInstanceIT.canSearchByTitle_partOfTitle:67 JSON path "totalRecords"
Expected: is <1>
     but: was <0>
[INFO] 
[ERROR] Tests run: 7, Failures: 2, Errors: 0, Skipped: 0
```

src/main/java/org/folio/search/service/converter/ElasticsearchHitConverter.java
```diff
@@ -0,0 +1,69 @@
...
+        .map(ElasticsearchHitConverter::processField)
+        .filter(Objects::nonNull)
+        .collect(Collectors.toList());
+    }
+    return isMultiLanguageField(value) ? getSourceMultilangValue(value) : value;
+  }
+
+  @SuppressWarnings("unchecked")
+  private static boolean isMultiLanguageField(Object fieldValue) {
+    return fieldValue instanceof Map && ((Map<String, Object>) fieldValue).containsKey("src");
```
MikhailFokanov, line 62: COULD you create CONSTATNT for string "src", as it is used more than once.

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
```diff
@@ -0,0 +1,170 @@
...
+    if (value instanceof Map) {
+      return convertMapUsingResourceFields((Map<String, Object>) value, subfields, ctx);
+    }
+
+    if (value instanceof List) {
+      return ((List<Object>) value).stream()
+        .map(listValue -> getObjectFieldValue(listValue, subfields, ctx))
+        .filter(Objects::nonNull)
+        .collect(toList());
+    }
```
bohdan-suprun, line 134: Will it work for plain string arrays (e.g. subjects)?

bohdan-suprun (approved):

PR #9, opened 2021-01-27
MSEARCH-9: Tenant level language support.

bohdan-suprun:
https://issues.folio.org/browse/MSEARCH-9 - Tenant level language support

## Changes:
* Introduce `search-config` API (with getAll, post, delete by id endpoints);
* Implement tenant init endpoint;
* Configure JPA and data migration via liquibase;
* Use configured languages during resource indexing;
* Implement unit tests.


src/main/java/org/folio/search/converter/LanguageConfigConverter.java
```diff
@@ -0,0 +1,27 @@
+package org.folio.search.converter;
+
+import java.util.UUID;
+import lombok.experimental.UtilityClass;
+import org.folio.search.domain.dto.LanguageConfig;
+import org.folio.search.model.config.LanguageConfigEntity;
+
+@UtilityClass
```
pfilippovEPAM, line 8: it is experimental feature, I guess in that case it can be replaced with just noArgConstructor

src/main/java/org/folio/search/service/SearchDocumentConverter.java
```diff
@@ -125,11 +126,13 @@ private static boolean isTextNode(JsonNode jsonNode) {
 
   private List<String> getResourceLanguages(String resourceName, DocumentContext doc) {
     var languageSourcePaths = descriptionService.getLanguageSourcePaths(resourceName);
+    var allLanguageCodes = languageConfigService.getAllSupportedLanguageCodes();
```
pfilippovEPAM, line 129:
```suggestion
    var supportedLanguages = languageConfigService.getAllSupportedLanguageCodes();
```

src/main/java/org/folio/search/service/LanguageConfigService.java
```diff
@@ -0,0 +1,65 @@
...
+import org.folio.search.exception.ValidationException;
+import org.folio.search.model.config.LanguageConfigEntity;
+import org.folio.search.repository.LanguageConfigRepository;
+import org.folio.search.service.metadata.ResourceDescriptionService;
+import org.springframework.stereotype.Service;
+
+@Log4j2
+@Service
+@AllArgsConstructor
+public class LanguageConfigService {
```
pfilippovEPAM, line 24: No unit tests for this class

src/main/java/org/folio/search/integration/KafkaMessageListener.java
```diff
@@ -39,7 +44,19 @@ public void handleEvents(List<ResourceEventBody> events) {
...
       .collect(toList());
 
-    indexService.indexResources(resources);
+    resources.stream()
+      .collect(groupingBy(ResourceEventBody::getTenant))
+      .forEach((tenant, eventsForTenant) -> {
+        try {
+          log.info("Processing [{}] events for tenant [{}]", eventsForTenant.size(), tenant);
+          // This needed to inject FOLIO specific values to properly work with DB
+          beginFolioExecutionContext(new AsyncFolioExecutionContext(tenant, moduleMetadata));
```
pfilippovEPAM, line 53: I guess this part should be tested in unit tests

src/main/java/org/folio/search/converter/LanguageConfigConverter.java
bohdan-suprun: Replaced with noArgsConstructor.

src/main/java/org/folio/search/service/LanguageConfigService.java
bohdan-suprun: Added, some of integration tests were removed and replaced with unit tests.

src/main/java/org/folio/search/integration/KafkaMessageListener.java
bohdan-suprun: good point, I've added a unit test for it.

src/main/java/org/folio/search/integration/KafkaMessageListener.java
```diff
@@ -39,7 +43,30 @@ public void handleEvents(List<ResourceEventBody> events) {
       .map(KafkaMessageListener::asInstanceResource)
       .collect(toList());
 
-    indexService.indexResources(resources);
+    resources.stream()
+      .collect(groupingBy(ResourceEventBody::getTenant))
```
pfilippovEPAM, line 47: I am not sure that splitting the batch into severals is a good idea, I am sure that there is an another way to retrieve tenant languages without creating FolioExecutionContexts

src/main/java/org/folio/search/model/config/LanguageConfigEntity.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@Table(name = "language_config")
+@Entity
+public class LanguageConfigEntity {
+  @Id
+  private UUID id;
```
pfilippovEPAM, line 18: why do we need uuid for the language entity, languageCode can be used as id for the table

bohdan-suprun: It is common approach on FOLIO I think, there is at least few more examples where a surrogate key is used instead of natural (locations, institutions (there is a code for each institution), etc.).

@MikhailFokanov, @hjiebsco should we proceed with this suggestion and get rid of UUID identifier and use natural primary key?

hjiebsco: lang code is cleaner in my view.

src/main/java/org/folio/search/integration/KafkaMessageListener.java
bohdan-suprun: I have modified this piece so that we only split events into batch for converting but indexing is still happens at once.

src/main/java/org/folio/search/model/config/LanguageConfigEntity.java
bohdan-suprun: Ok, I'll use then language code as PK.

bohdan-suprun: Igor has confirmed that it is fine from UI point of view do not have UUID. The UUID has been removed.

hjiebsco (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -41,6 +41,44 @@
...
+    {
+      "id" : "_tenant",
+      "version" : "2.0",
+      "interfaceType": "system",
+      "handlers": [
+        {
+          "methods": ["POST"],
+          "pathPattern": "/_/tenant"
+        }
+      ]
```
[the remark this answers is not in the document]
bohdan-suprun: I have created https://issues.folio.org/browse/MSEARCH-45 for this, we have to update folio-spring-base module so it should propagate x-okapi-tenant header.

I also downgraded tenant API version to 1.2 because we don't support the new async style.

src/test/java/org/folio/search/controller/ConfigControllerIT.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  @Test
+  void shouldUseConfiguredLanguagesDuringMapping() {
+    final List<String> languageCodes = List.of("eng", "rus");
+    for (String languageCode : languageCodes) {
+      doPost(languageConfig(), new LanguageConfig().code(languageCode));
+    }
+
+    var newInstance = InstanceBuilder.builder()
```
pfilippovEPAM, line 87: there is an Instance object, can't we use it?

src/main/resources/swagger.api/schemas/common/error.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
pfilippovEPAM, line 2: I am not sure that is the part of PR, look at the MSEARCH-39

src/main/java/org/folio/search/controller/ErrorHandler.java
```diff
@@ -0,0 +1,23 @@
...
+import org.folio.search.exception.ValidationException;
+import org.folio.tenant.domain.dto.Parameter;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+
+@RestControllerAdvice
+public class ErrorHandler {
+  @ExceptionHandler(ValidationException.class)
+  public ResponseEntity<ValidationErrors> handleValidationException(ValidationException ex) {
```
pfilippovEPAM, line 16: this part of PR related to msearch-39

src/test/java/org/folio/search/repository/IndexRepositoryTest.java
```diff
@@ -78,7 +78,7 @@ void createIndex_negative_throwsException() throws IOException {
       .isInstanceOf(SearchServiceException.class)
       .hasCauseExactlyInstanceOf(IOException.class)
       .hasMessage("Failed to perform elasticsearch request "
-        + "[index=test-resource_test-tenant, type=createIndexApi, message: err]");
```
pfilippovEPAM, line 81: what the reason of this change?

src/test/java/org/folio/search/service/converter/SearchDocumentConverterTest.java
```diff
@@ -54,12 +68,12 @@ void convertSingle_positive() {
     var resourceDescription = resourceDescription(getDescriptionFields(), List.of("$.language"));
 
     when(descriptionService.get(RESOURCE_NAME)).thenReturn(resourceDescription);
-    when(descriptionService.isSupportedLanguage("eng")).thenReturn(true);
+    when(languageConfigService.getAllSupportedLanguageCodes()).thenReturn(Set.of("eng"));
 
-    var actual = documentMapper.convert(eventBody);
+    var actual = documentMapper.convert(List.of(eventBody));
```
pfilippovEPAM, line 73: name of method is convertSingle, why list of values passed?

src/test/java/org/folio/search/service/converter/SearchDocumentConverterTest.java
```diff
@@ -89,10 +103,10 @@ void convertSingle_negative_pathNotFound() {
 
     when(descriptionService.get(RESOURCE_NAME)).thenReturn(resourceDescription);
 
-    var actual = documentMapper.convert(eventBody);
+    var actual = documentMapper.convert(List.of(eventBody));
```
pfilippovEPAM, line 106: the same as above

src/test/java/org/folio/search/service/converter/SearchDocumentConverterTest.java
```diff
@@ -126,12 +140,12 @@ void convertSingle_positive_multilangResource() {
       List.of("$.lang1", "$.lang2", "$.lang3", "$.lang4", "$.lang5"));
 
     when(descriptionService.get(RESOURCE_NAME)).thenReturn(resourceDescription);
-    when(descriptionService.isSupportedLanguage("eng")).thenReturn(true);
+    when(languageConfigService.getAllSupportedLanguageCodes()).thenReturn(Set.of("eng"));
 
-    var actual = documentMapper.convert(eventBody);
+    var actual = documentMapper.convert(List.of(eventBody));
```
pfilippovEPAM, line 145: the same as above

src/test/java/org/folio/search/repository/IndexRepositoryTest.java
bohdan-suprun: The tenant name is used to create postgresql schema, but `test-tenant` is not a valid identifier, so I had to change it to `test_tenant`.

src/main/java/org/folio/search/controller/ErrorHandler.java
bohdan-suprun: Sure, but I have some tests that verifies that validation error is returned, they will fail if I remove this code.

src/test/java/org/folio/search/controller/ConfigControllerIT.java
bohdan-suprun: Yes, the builder is removed.

src/test/java/org/folio/search/service/converter/SearchDocumentConverterTest.java
bohdan-suprun: The test method is renamed. 

The `convert(ResourceEventBody)` has `private` access now because it's signature has been changed to `convert(IndexingConfig, ResourceEventBody)`

pfilippovEPAM: So, I assume, indexing config should be passed to this method instead of changing method privacy and testName?

src/main/java/org/folio/search/controller/ErrorHandler.java
pfilippovEPAM: ok, Also I have a question, why do we need custom ValidationException, when Spring and Java provide other options? like IllegalArgumentException or ConstraintViolationException

src/test/java/org/folio/search/repository/IndexRepositoryTest.java
pfilippovEPAM: This should be discussed, maybe I missed some information about tenant name requirements, it should be always in lower snake case?

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
```diff
@@ -76,15 +94,39 @@
...
       .distinct()
-      .filter(descriptionService::isSupportedLanguage)
+      .filter(indexingConfig::isSupportedLanguage)
       .collect(toList());
   }
 
+  private Stream<SearchDocumentBody> convertForTenant(String tenant, List<ResourceEventBody> events) {
+    try {
+      beginFolioExecutionContext(tenant);
+      final var indexingConfig = new IndexingConfig(languageConfigService.getAllSupportedLanguageCodes());
```
pfilippovEPAM, line 111: why do we need here final modifier?

src/test/java/org/folio/search/repository/IndexRepositoryTest.java
bohdan-suprun: Case is not sensitive here, but it should be a valid postgres identifier I think.

src/main/java/org/folio/search/controller/ErrorHandler.java
bohdan-suprun: I'm fine using existing exceptions. What exception are you suggesting for validation? `ConstraintViolationException` looks quite tricky to construct in order to be able to return the FOLIO standard validation error response, i.e.:
```
{
  "message": "..."
  "parameters": [{
     "key": "the caused property name",
     "value": "the value caused the issue"
  }]
}
```

src/test/java/org/folio/search/service/converter/SearchDocumentConverterTest.java
bohdan-suprun: This is now reverted.

bohdan-suprun:
Hi @pfilippovEPAM, @MikhailFokanov, This PR is quite long in review, could one of you review and approve it?

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
```diff
@@ -39,52 +41,55 @@
...
-
-    var conversionContext = ConversionContext.of(
-      getResourceLanguages(resourceData, resourceDescription.getLanguageSourcePaths()));
-    Map<String, Object> resultDocument = convertMapUsingResourceFields(resourceData, fields, conversionContext);
-
-    return Optional.of(SearchDocumentBody.builder()
-      .id(MapUtils.getString(resourceData, "id"))
-      .index(getElasticsearchIndexName(event.getResourceName(), event.getTenant()))
-      .routing(event.getTenant())
+  private SearchDocumentBody convert(ConversionContext context) {
```
pfilippovEPAM, line 59: what the reason of changing method signature?

pfilippovEPAM (approved):

PR #10, opened 2021-01-28
MSEARCH-18 Implement search by publishers

pfilippovEPAM:
- Implement source fields
- Update swagger documentation
- Update integration tests

### DIFF
--- a/src/main/java/org/folio/search/service/metadata/SearchFieldProvider.java
+++ b/src/main/java/org/folio/search/service/metadata/SearchFieldProvider.java
@@ -20,9 +20,17 @@ public interface SearchFieldProvider {
   /**
    * Provides list of fields for given search type.
    *
-   * @param searchType search type as {@link String}
    * @param resource resource type as {@link String}
+   * @param searchType search type as {@link String}
    * @return list of fields.
    */
   List<String> getFields(String resource, String searchType);
+
+  /**
+   * Provides list of fields of source fields for resource.
+   *
+   * @param resource resource type as {@link String}
+   * @return list of fields.
+   */
+  List<String> getSourceFields(String resource);
 }
--- a/src/main/resources/model/instance.json
+++ b/src/main/resources/model/instance.json
@@ -95,6 +95,20 @@
           "index": "keyword"
         }
       }
+    },
+    "publication": {
+      "type": "object",
+      "properties": {
+        "publisher": {
+          "inventorySearchTypes": "publisher",
+          "index": "multilang",
+          "showInResponse": true
+        },
+        "dateOfPublication": {
+          "index": "source",
+          "showInResponse": true
+        }
+      }
     }
   },
   "groups": { },
--- a/pom.xml
+++ b/pom.xml
@@ -196,7 +196,14 @@
       <artifactId>h2</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>3.0.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
+
 
   <build>
     <plugins>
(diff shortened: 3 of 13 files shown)


src/main/resources/model/instance.json
```diff
@@ -95,6 +95,20 @@
...
+    "publication": {
+      "type": "object",
+      "properties": {
+        "publisher": {
+          "inventorySearchTypes": "publisher",
+          "index": "multilang",
+          "showInResponse": true
+        },
+        "dateOfPublication": {
+          "index": "source",
```
bohdan-suprun, line 108: What is `source` index type means?

src/main/resources/model/instance.json
```diff
@@ -95,6 +95,20 @@
           "index": "keyword"
         }
       }
+    },
+    "publication": {
+      "type": "object",
+      "properties": {
+        "publisher": {
+          "inventorySearchTypes": "publisher",
```
bohdan-suprun, line 103: does it mean that user should use `publisher=="abcd"` CQL query?

src/main/java/org/folio/search/service/metadata/SearchFieldProvider.java
```diff
@@ -20,9 +20,17 @@
...
    * @return list of fields.
    */
   List<String> getFields(String resource, String searchType);
+
+  /**
+   * Provides list of fields of source fields for resource.
+   *
+   * @param resource resource type as {@link String}
+   * @return list of fields.
+   */
```
bohdan-suprun, line 34: I don't think this comment adds more information, I'd prefer to remove it if that's fine.

bohdan-suprun (commented):
Is there an integration test for the search?

src/main/resources/model/instance.json
pfilippovEPAM: it means that value would be stored in elasticsearch, but it would be able for search (ES won't create reverse index for it)

src/main/resources/model/instance.json
pfilippovEPAM: yes. that's right. I don't find search by publication in existing inventory search, but I guess it would be better to shorten name of field

src/main/java/org/folio/search/service/metadata/SearchFieldProvider.java
pfilippovEPAM: it's default javadoc structure - Header of javadoc, @param - for each parameter of method and return value.

src/main/resources/model/instance.json
bohdan-suprun: Yeah, shortening sound good for me, however it can be not so clear for users (most of them are expecting that properties matches json schema). We should document it somewhere (for example in module readme) if we're going to procced with the custom name.

 What do you think?

src/main/java/org/folio/search/service/metadata/SearchFieldProvider.java
bohdan-suprun: Sure, AFAIK on FOLIO it is not required to have comments if method name is clear enough.

pfilippovEPAM: but it does not bring any issues in further, anyone can extend javadoc when implementation become more complicated or will require more detailed description

From perspective of checkstyle - javadoc is optional, but my IDE is set up to generate default stubs for any public method, so I just fill required fields in this template

bohdan-suprun: Usually folks (including me) forgot to update java docs :(. But that's fine, let's leave them as they are.

src/main/resources/model/instance.json
pfilippovEPAM: I suppose that we can postpone a little bit a documenting of all features until current sprint search features will be completed and then create a single PR with whole options of search - for example: search by contributors looks like in CQL like ```(contributors =/@name "123") sortby title```

bohdan-suprun (approved):

hjiebsco (approved):

PR #11, opened 2021-01-29
MSEARCH-17: Allow multilang search for contributors.name property.

bohdan-suprun:
https://issues.folio.org/browse/MSEARCH-17 - Instance - support contributors keyword search

### DIFF
--- a/src/main/resources/model/instance.json
+++ b/src/main/resources/model/instance.json
@@ -54,7 +54,8 @@
       "type": "object",
       "properties": {
         "name": {
-          "index": "keyword",
+          "inventorySearchTypes": "contributors.name",
+          "index": "multilang",
           "showInResponse": true
         }
       }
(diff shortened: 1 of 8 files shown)


src/main/resources/model/instance.json
```diff
@@ -54,7 +54,8 @@
       "type": "object",
       "properties": {
         "name": {
-          "index": "keyword",
+          "inventorySearchTypes": "contributors.name",
```
pfilippovEPAM, line 57: existing contract for query - ```(contributors =/@name "harry") sortby title```, according to this change - UI should change query to ```(contributors.name ="harry1") sortby title````, but we agreed that we support existing contract 1 to 1

hjiebsco (approved):

src/main/resources/model/instance.json
bohdan-suprun: Now we support both notations

pfilippovEPAM (approved):

PR #13, opened 2021-02-02
MSEARCH-15, MSEARCH-19: Support subjects and keyword search.

bohdan-suprun:
https://issues.folio.org/browse/MSEARCH-15 - Instance - support keyword search fields - (title, contributors, identifiers)
https://issues.folio.org/browse/MSEARCH-19 - Instance - support keyword search subjects field

## Changes:
* Make sure `keyword` inventory search includes:
   * title (string)
   * alternativeTitles (alternativeTitles[].alternativeTitle)
   * indexTitle (string)
   * series (array of strings)
   * contributors (contributors[].name
   * identifiers (all) (identifiers[].value) - see note 1 below
* Support multilang search by `subjects` property;
* Add tests.


src/main/resources/model/instance.json
```diff
@@ -43,6 +43,7 @@
       "type": "object",
       "properties": {
         "value": {
+          "inventorySearchTypes": "title",
```
pfilippovEPAM, line 46: why identifiers start related to the title search? 
title search is finished in https://issues.folio.org/browse/MSEARCH-16

src/main/resources/model/instance.json
```diff
@@ -54,14 +55,14 @@
       "type": "object",
       "properties": {
         "name": {
-          "inventorySearchTypes": "contributors",
+          "inventorySearchTypes": ["contributors", "title"],
```
pfilippovEPAM, line 58: the same as above, contributors is not a part of title search

src/test/java/org/folio/search/service/metadata/LocalSearchFieldProviderTest.java
```diff
@@ -85,6 +87,18 @@ void shouldAddStarNotationForMultilangField() {
       .containsExactly("title2.sub3.sub5.*");
   }
 
+  @Test
+  void shouldReturnAllExpectedFieldsForInstanceTitleSearch() {
```
pfilippovEPAM, line 91: this test looks like a redundant:
1. It depends of existing model, which can be modified
2. Behavior of inventory search types already tested in other unit tests

src/main/resources/model/instance.json
bohdan-suprun: Thanks, it should be `keyword` instead.

src/test/java/org/folio/search/service/metadata/LocalSearchFieldProviderTest.java
bohdan-suprun: This test has been removed.

hjiebsco (approved):

src/main/resources/model/instance.json
```diff
@@ -26,23 +26,24 @@
...
     "series": {
-      "inventorySearchTypes": "title",
+      "inventorySearchTypes": ["title", "keyword"],
       "index": "multilang"
     },
     "identifiers": {
       "type": "object",
       "properties": {
         "value": {
+          "inventorySearchTypes": ["title", "keyword"],
```
pfilippovEPAM, line 46: identifiers should be only keyword, without title search group

bohdan-suprun: thanks, removed.

PR #14, opened 2021-02-02
MSEARCH-39 Implement exception handling for rest controllers

pfilippovEPAM:


src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Catches and handles all {@link SearchServiceException} objects during code execution.
+   *
+   * @param exception {@link SearchServiceException} to process
+   * @return {@link ResponseEntity} with {@link ErrorResponse} body
+   */
+  @ExceptionHandler(SearchServiceException.class)
+  public ResponseEntity<ErrorResponse> handleSearchServiceException(SearchServiceException exception) {
+    log.debug("Handling exception [type: SearchServiceException]", exception);
```
hjiebsco, line 36: Do you have to hardcode the `type: SearchServiceException`? Maybe `exception.getClass().getSimpleName()` or just `exception` will be enough.

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,148 @@
...
+    var defaultErrorResponse = new ErrorResponse()
+      .addErrorsItem(new Error()
+        .message(exception.getMessage())
+        .type(SearchServiceException.class.getSimpleName())
+        .code(ErrorCode.SERVICE_ERROR.getDescription()))
+      .totalRecords(1);
+
+    return ResponseEntity.status(BAD_REQUEST).body(defaultErrorResponse);
```
hjiebsco, line 44: Does `SearchServiceException` always have the `ErrorCode.SERVICE_ERROR`? If yes, maybe you can add it to the exception class. Also this code block seems to very similar to Line 62-67. Consider simplify and consolidate it.

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,148 @@
...
+    if (cause instanceof ElasticsearchException) {
+      return ResponseEntity.badRequest().body(buildErrorResponse((ElasticsearchException) cause));
+    }
+
+    return ResponseEntity.badRequest().body(new ErrorResponse()
+      .addErrorsItem(new Error()
+        .message(exception.getMessage())
+        .type(SearchOperationException.class.getSimpleName())
+        .code(ErrorCode.UNKNOWN_ERROR.getDescription()))
+      .totalRecords(1));
```
hjiebsco, line 67: Similar to above comments.

src/main/java/org/folio/search/model/types/ErrorType.java
```diff
@@ -0,0 +1,16 @@
...
+import com.fasterxml.jackson.annotation.JsonValue;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public enum ErrorType {
+
+  INTERNAL("-1"),
+  UNKNOWN("-4");
```
hjiebsco, line 12: The use of number feels like this class is more like `ErrorCode` than `ErrorType`

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
bohdan-suprun, line 36: Should it be warn or at least info?

src/main/resources/swagger.api/examples/errors.sample
```diff
@@ -0,0 +1,15 @@
+{
+  "errors": [
+    {
+      "message": "must not be null",
+      "type": "-1",
+      "code": "Validation error",
```
hjiebsco, line 6: Type has number but code has string. Feels a bit the opposite.

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,148 @@
...
+   * @return {@link ResponseEntity} with {@link ErrorResponse} body
+   */
+  @ExceptionHandler(SearchServiceException.class)
+  public ResponseEntity<ErrorResponse> handleSearchServiceException(SearchServiceException exception) {
+    log.debug("Handling exception [type: SearchServiceException]", exception);
+    var defaultErrorResponse = new ErrorResponse()
+      .addErrorsItem(new Error()
+        .message(exception.getMessage())
+        .type(SearchServiceException.class.getSimpleName())
+        .code(ErrorCode.SERVICE_ERROR.getDescription()))
```
bohdan-suprun, line 41: Does it make sense to make the ErrorCode specific to the area: e.g. SEARCH_ERROR, INDEXING_ERROR, etc?

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,148 @@
...
+   * Catches and handles all {@link SearchOperationException} objects during code execution.
+   *
+   * @param exception {@link SearchOperationException} to process
+   * @return {@link ResponseEntity} with {@link ErrorResponse} body
+   */
+  @ExceptionHandler(SearchOperationException.class)
+  public ResponseEntity<ErrorResponse> handleSearchServiceException(SearchOperationException exception) {
+    log.warn("Handling exception [type: SearchOperationException]", exception);
+
+    var cause = exception.getCause();
```
bohdan-suprun, line 57: Should we walk through the whole stack just in case?

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,148 @@
...
+    if (cause instanceof ElasticsearchException) {
+      return ResponseEntity.badRequest().body(buildErrorResponse((ElasticsearchException) cause));
```
bohdan-suprun, line 59: Why do we know that it is bad request but not internal server error?

src/main/resources/swagger.api/mod-search.yaml
```diff
@@ -25,6 +25,20 @@ paths:
               example: examples/searchResult.sample
               schema:
                 $ref: '#/components/schemas/searchResult'
+        '400':
+          description: Error response when NullPointerException occurred during code execution
```
hjiebsco, line 29: This description seems to describe only one specific cause.

src/main/resources/swagger.api/mod-search.yaml
```diff
@@ -25,6 +25,20 @@ paths:
...
+          content:
+            application/json:
+              example: examples/indexNotFoundError.sample
+              schema:
+                $ref: '#/components/schemas/errorResponse'
+        '500':
+          description: Error response when NullPointerException occurred during code execution
```
hjiebsco, line 36: Same as above. The description describes only one cause.

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,148 @@
...
+   * Catches and handles all exceptions of type {@link MethodArgumentNotValidException}.
+   *
+   * @param e {@link MethodArgumentNotValidException} to process
+   * @return {@link ResponseEntity} with {@link ErrorResponse} body
+   */
+  @ExceptionHandler(MethodArgumentNotValidException.class)
+  public ResponseEntity<ErrorResponse> handleMethodArgumentNotValidException(MethodArgumentNotValidException e) {
+    log.debug("handling exception [type: MethodArgumentNotValidException]", e);
+    var errorResponse = new ErrorResponse();
+    e.getBindingResult().getAllErrors().forEach(error ->
```
bohdan-suprun, line 80: Is it null safe?

src/main/java/org/folio/search/model/types/ErrorType.java
```diff
@@ -0,0 +1,16 @@
+package org.folio.search.model.types;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public enum ErrorType {
```
hjiebsco, line 9: Could not find reference in the project. Is this used anywhere?

src/main/resources/swagger.api/mod-search.yaml
bohdan-suprun, line 29: `Error response when NullPointerException occurred during code execution` shouldn't it be a 500 error?

src/main/resources/swagger.api/mod-search.yaml
```diff
@@ -45,6 +59,20 @@ paths:
...
                 $ref: '#/components/schemas/folioCreateIndexResponse'
+        '400':
+          description: Error response when request tries to create index that already exists
+          content:
+            application/json:
+              example: examples/indexCreationError.sample
+              schema:
+                $ref: '#/components/schemas/errorResponse'
+        '500':
+          description: Error response when NullPointerException occurred during code execution
```
bohdan-suprun, line 70: There might be other possible errors, but not only NPE, should we update description to something:
 ```suggestion
          description: When unhandled exception occurred during code execution, e.g. NullPointerException
```

File: src/test/java/org/folio/search/controller/IndexControllerTest.java [hunk 14]
```diff
@@ -52,6 +58,102 @@ void createIndex_positive() throws Exception {
...
+
+    when(indexService.createIndex(RESOURCE_NAME, TENANT_ID)).thenThrow(
+      new SearchOperationException("error", elasticsearchException));
+
+    mockMvc.perform(requestBuilder)
+      .andExpect(status().isBadRequest())
+      .andExpect(jsonPath("$.errors[0].message", is("Index already exists: " + INDEX_NAME)))
+      .andExpect(jsonPath("$.errors[0].type", is("ElasticsearchException")))
+      .andExpect(jsonPath("$.errors[0].code", is("Elasticsearch error")))
+      .andExpect(jsonPath("$.total_records", is(1)));
```
bohdan-suprun, line 81: Can it be the first assertion?

src/test/java/org/folio/search/controller/IndexControllerTest.java
```diff
@@ -52,6 +58,102 @@ void createIndex_positive() throws Exception {
       .andExpect(jsonPath("$.indices", is(List.of(INDEX_NAME))));
   }
 
+  @Test
+  void createIndex_negative_indexAlreadyExists() throws Exception {
+    var requestBuilder = post("/search/index/indices")
```
bohdan-suprun, line 63: Could you extract it to a constant or put it to `org.folio.search.support.base.ApiEndpoints` class, so that it can be reused?

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
pfilippovEPAM: This exception can occur frequently - for example when user provides invalid CQL search query, I don't see the reason to log every validation error at info or warn level

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
pfilippovEPAM: No, SearchOperationException wraps all elasticsearch level exceptions from Rest high level client

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
pfilippovEPAM: In the most cases this error will be used to wrap uncaught exceptions from elasticsearch - like invalid mappings or index does not exists and so on - it will be bad request, not internal server error, other cases when elasticsearch failed or stopped - it should have different level of status, but I am not sure that we will face with it at the POC stage, in future this class can be easily extended with new cases (if would be required)

src/test/java/org/folio/search/controller/IndexControllerTest.java
pfilippovEPAM: why? any of this assertion will fail if response object does not contain correct value at the json path value, what order of assertions can change test behaviour?

src/main/java/org/folio/search/model/types/ErrorType.java
pfilippovEPAM: Looks like I have copied this class from other service, I will remove it

src/test/java/org/folio/search/controller/IndexControllerTest.java
bohdan-suprun: It should be a little bit clearer to find out cause if test is failed. 

If `errors` is empty, I assume the assertion will fail with something like: expected `Index already exists: instance` but actual `null` and it is not clear whether there were no message or errors is empty.

src/main/resources/swagger.api/examples/errors.sample
pfilippovEPAM: I'll removed outdated functionality and ErrorType enum, because I copied them by mistake

src/main/java/org/folio/search/exception/SearchServiceException.java
```diff
@@ -1,10 +1,15 @@
...
 
+import lombok.Getter;
+import org.folio.search.model.types.ErrorCode;
+
 /**
  * Thrown to indicate a service error that was occurred at operations to search engine.
  */
 public class SearchServiceException extends RuntimeException {
 
+  @Getter private final ErrorCode errorCode = ErrorCode.SERVICE_ERROR;
```
bohdan-suprun, line 11: Maybe we should introduce a base exception with the property and then all our custom exceptions have to set the error code?

bohdan-suprun (approved):

src/main/java/org/folio/search/controller/ApiExceptionHandler.java
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * Handles all uncaught exceptions.
+   *
+   * @param exception {@link Exception} object
+   * @return {@link ResponseEntity} with {@link ErrorResponse} body.
+   */
+  @ExceptionHandler(Exception.class)
+  public ResponseEntity<ErrorResponse> handleAllOtherExceptions(Exception exception) {
+    log.warn("Handling exception [type: {}]", exception.getClass().getSimpleName(), exception);
```
hjiebsco, line 113: For consistence, maybe just `log.warn("Handling exception", exception);`. The stack trace will contain exception class name.

src/main/java/org/folio/search/model/types/ErrorCode.java
```diff
@@ -0,0 +1,17 @@
...
+@RequiredArgsConstructor
+public enum ErrorCode {
+
+  UNKNOWN_ERROR("Unknown error"),
+  SERVICE_ERROR("Service error"),
+  ELASTICSEARCH_ERROR("Elasticsearch error"),
+  VALIDATION_ERROR("Validation error"),
+  CONSTRAINT_VIOLATION("Constraint violation error");
+
+  private final String description;
```
hjiebsco, line 16: Error codes are more like constants. I am wondering if description is needed.

src/main/resources/swagger.api/schemas/error.json
```diff
@@ -0,0 +1,26 @@
...
+      "description": "Error message code"
+    },
+    "parameters": {
+      "type": "object",
+      "description": "Error message parameters",
+      "$ref": "parameters.json"
+    }
+  },
+  "additionalProperties": false,
+  "required": [ "message" ]
```
hjiebsco, line 25: Is only `message` required, not `code` and/or `type`? Also earlier `nptError.sample` has only `type` and `code`, but no `message`.

hjiebsco (commented):
If clients want to take action based on the error mod-search reported (error.json is a contract), what info will be sufficient but do not expose too much? I am wondering if both `type` and `code` are needed. Also I am thinking that `code` should be more like a constant (with understore). What do you think?

pfilippovEPAM:
> If clients want to take action based on the error mod-search reported (error.json is a contract), what info will be sufficient but do not expose too much? I am wondering if both `type` and `code` are needed. Also I am thinking that `code` should be more like a constant (with understore). What do you think?

I am agree with that, I will replace code phrases with real codes like `elasticsearch_error`,

src/main/resources/swagger.api/schemas/error.json
pfilippovEPAM: Yes, for NPE errors java does not provide by default error message, but this is rare case, that should not be occurred in production. And I suppose this validation only works in request body, but not in response

hjiebsco (approved):

PR #15, opened 2021-02-03
MSEARCH-2 Implement integration tests for hrid search

pfilippovEPAM:


src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -2,119 +2,63 @@
...
+      arguments("search by instance title (alternative title)", "title all {value}", array("primers"), null),
+      arguments("search by publisher (abbreviate)", "publisher all {value}", array("MIT"), null),
+      arguments("search by publisher (abbreviate lowercase)", "publisher all {value}", array("mit"), null),
+      arguments("search by publisher (word)", "publisher all {value}", array("press"), null),
+      arguments("search by contributors name", "contributors.name all {value}", array("frank"),
+        (ThrowingConsumer<ResultActions>) searchResult -> searchResult
+          .andExpect(jsonPath("instances[0].contributors[0].name", is("Antoniou, Grigoris")))
+          .andExpect(jsonPath("instances[0].contributors[1].name", is("Van Harmelen, Frank")))),
+      arguments("search by contributors alias", "contributors all {value}", array("grigoris"), null),
+      arguments("search by hrid for exact match", "hrid={value}", array("inst000000000022"), null),
```
bohdan-suprun, line 58: CQL uses `==` for exact match.

src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -2,119 +2,63 @@
...
+      arguments("search by instance title (series, partial)", "title all {value}", array("cooperate"), null),
+      arguments("search by instance title (part of title)", "title all {value}", array("primers"), null),
+      arguments("search by instance title (alternative title)", "title all {value}", array("primers"), null),
+      arguments("search by publisher (abbreviate)", "publisher all {value}", array("MIT"), null),
+      arguments("search by publisher (abbreviate lowercase)", "publisher all {value}", array("mit"), null),
+      arguments("search by publisher (word)", "publisher all {value}", array("press"), null),
+      arguments("search by contributors name", "contributors.name all {value}", array("frank"),
+        (ThrowingConsumer<ResultActions>) searchResult -> searchResult
+          .andExpect(jsonPath("instances[0].contributors[0].name", is("Antoniou, Grigoris")))
+          .andExpect(jsonPath("instances[0].contributors[1].name", is("Van Harmelen, Frank")))),
```
bohdan-suprun, line 56: Feel free to remove it, It is covered by a separate test.

src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -2,119 +2,63 @@
...
+      arguments("search by contributors alias", "contributors all {value}", array("grigoris"), null),
+      arguments("search by hrid for exact match", "hrid={value}", array("inst000000000022"), null),
+      arguments("search by hrid with wildcard (starts with)", "hrid={value}", array("inst000*"), null),
+      arguments("search by hrid with wildcard (ends with)", "hrid={value}", array("*00022"), null),
+      arguments("search by hrid with wildcard (contains)", "hrid={value}", array("*00000002*"), null)
```
bohdan-suprun, line 61: the 3rd argument is array, but all the cases requires only one argument. Can we make it a single value?

bohdan-suprun (approved):

hjiebsco (approved):

src/test/java/org/folio/search/controller/SearchInstanceIT.java
pfilippovEPAM: yes, but if we will implement bool query with 2 arguments, we have to refactor this again

src/test/java/org/folio/search/controller/SearchInstanceIT.java
pfilippovEPAM: I will leave it as example for future development, this check does not takes a lot of time, but can be handful in future cases with sorting and so on

PR #16, opened 2021-02-03
MSEARCH-46: Add expandAll property for /search/instances

bohdan-suprun:
https://issues.folio.org/browse/MSEARCH-46 - Allow returning of all instance properties.

## Changes:
* Add new `expandAll` property for the `/search/instances` endpoint;
* Bump up minor version for the API;
* Add tests for the new change.
* Make `identifierTypeId` as source.

### DIFF
--- a/src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
+++ b/src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
@@ -74,9 +74,16 @@ public class CqlSearchQueryConverter {
       }
     }
 
+    if (!searchRequest.isExpandAll()) {
+      final String[] includes = searchFieldProvider
+        .getSourceFields(searchRequest.getResource())
+        .toArray(String[]::new);
+
+      queryBuilder.fetchSource(includes, null);
+    }
+
     return queryBuilder
       .query(convertToQuery(searchRequest, node))
-      .fetchSource(searchFieldProvider.getSourceFields(searchRequest.getResource()).toArray(String[]::new), null)
       .from(searchRequest.getOffset())
       .size(searchRequest.getLimit());
   }
--- a/src/main/resources/model/instance.json
+++ b/src/main/resources/model/instance.json
@@ -47,7 +47,7 @@
           "index": "keyword"
         },
         "identifierTypeId": {
-          "index": "none"
+          "index": "source"
         }
       }
     },
--- a/src/test/java/org/folio/search/controller/EsInstanceToInventoryInstanceIT.java
+++ b/src/test/java/org/folio/search/controller/EsInstanceToInventoryInstanceIT.java
@@ -34,4 +34,17 @@ class EsInstanceToInventoryInstanceIT extends BaseIntegrationTest {
     assertThat(actual.getContributors(), is(expected.getContributors()));
     assertThat(actual.getPublication(), is(expected.getPublication()));
   }
+
+  @Test
+  void responseContainsAllInstanceProperties() throws Exception {
+    final var expected = getSemanticWeb();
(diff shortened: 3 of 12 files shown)


hjiebsco (approved):

src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
```diff
@@ -74,9 +74,16 @@ private SearchSourceBuilder toCriteria(CqlSearchRequest searchRequest, CQLNode n
       }
     }
 
+    if (!searchRequest.isExpandAll()) {
+      final String[] includes = searchFieldProvider
```
pfilippovEPAM, line 78: why do you need final modifier here?

src/main/resources/model/instance.json
```diff
@@ -47,7 +47,7 @@
           "index": "keyword"
         },
         "identifierTypeId": {
-          "index": "none"
+          "index": "source"
```
pfilippovEPAM, line 50: this change is not relevant to the expandAll functionality, is this preparation for some story?

src/test/java/org/folio/search/controller/EsInstanceToInventoryInstanceIT.java
```diff
@@ -34,4 +34,17 @@ void responseContainsOnlyBasicInstanceProperties() throws Exception {
     assertThat(actual.getContributors(), is(expected.getContributors()));
     assertThat(actual.getPublication(), is(expected.getPublication()));
   }
+
+  @Test
+  void responseContainsAllInstanceProperties() throws Exception {
+    final var expected = getSemanticWeb();
+    final var actualJson = mockMvc.perform(get(searchInstancesByQuery("id=={value}&expandAll=true"), expected.getId())
```
pfilippovEPAM, line 41: the same with final word, is this some kind code style?

src/main/resources/model/instance.json
bohdan-suprun: I needed it for testing to be able to use equals assertion, but this identifier should be also required for ISBN normalized stories, so I just decided to add it to index.

src/test/java/org/folio/search/controller/EsInstanceToInventoryInstanceIT.java
bohdan-suprun: Nope, just my personal preference.

src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
bohdan-suprun: See my response below

src/test/java/org/folio/search/controller/EsInstanceToInventoryInstanceIT.java
pfilippovEPAM: It looks redundant here and in the other occurrences, it makes sense in code only if you really want to show that this field value should not be changed, otherwise - it is overloading of code by using redundant modifiers, other developers know, that there is no need to rewrite value of method parameter

hjiebsco: Back to the old days, the best practices mentioned that `final` helps to indicate coder's intention and also helps with potential complier optimization. But for a very local scoped variables, it is not necessary. Simpler/cleaner code style is more preferred in my view.

pfilippovEPAM (approved):

src/test/java/org/folio/search/controller/EsInstanceToInventoryInstanceIT.java
bohdan-suprun: My intention was that a `final` variable must be never changed, even by mistake.

In this case the `expected` value should be constant and never changed. But i'm fine to not put it, if it makes code clearer.

PR #17, opened 2021-02-04
MSEARCH-20 Implement search by fields

pfilippovEPAM:
- electronicAccess [(MSEARCH-22)](https://issues.folio.org/browse/MSEARCH-22)
- classification [(MSEARCH-20)](https://issues.folio.org/browse/MSEARCH-20)


src/main/resources/model/instance.json
```diff
@@ -111,6 +111,42 @@
...
+      "properties": {
+        "note": {
+          "index": "multilang"
+        },
+        "staffOnly": {
+          "index": "bool"
+        }
+      }
+    },
+    "electronicAccess": {
```
bohdan-suprun, line 134: Should we add an 'alias' for all the subfields? So that we will be able to search by `electronicsAccess all "..."`?

In inventory we do support such a search https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/schema.json#L676

src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -61,7 +61,23 @@ void canSearchByInstances(String testName, String query, Object[] queryArguments
...
-      arguments("search by subjects", "subjects all {value}", array("semantic"), null)
+      arguments("search by subjects", "subjects all {value}", array("semantic"), null),
+      arguments("search by classification number",
+        "classifications.classificationNumber=={value}", array("025.04"), null),
+      arguments("search by classification number (using wildcard)",
+        "classifications.classificationNumber=={value}", array("025*"), null),
+      arguments("search by notes (word in note)", "notes.note all {value}", array("references"), null),
+      arguments("search by notes (partial match)", "notes.note all {value}", array("reference"), null),
+      arguments("search by notes (root of word)", "notes.note all {value}", array("readability"), null),
+      arguments("search by staffOnly", "notes.staffOnly=={value}", array("false"), null),
```
bohdan-suprun, line 72: I recall Magda asked if it is possible to implement search by public notes only (i.e. `staffOnly` is `false`), it is not possible, right?

hjiebsco (commented):
I second @bohdan-suprun's comments. Can we get clarification from @magdazacharska about [MSEARCH-21](https://issues.folio.org/browse/MSEARCH-21) if `staff-only` notes are allowed to be indexed but not searchable, or neither?, How to we control that?

bohdan-suprun (approved):

pfilippovEPAM:
> I second @bohdan-suprun's comments. Can we get clarification from @magdazacharska about MSEARCH-21 if staff-only notes are allowed to be indexed but not searchable, or neither?, How do we control that?

After discussion with @magdazacharska - I will implement new approach to search only by not staffOnly notes, but it will be done in a separate PR

hjiebsco (approved):

PR #18, opened 2021-02-08
MSEARCH-40: Add sort title and contributors indexes.

bohdan-suprun:
https://issues.folio.org/browse/MSEARCH-40 - Sorting results - title and contributor.

## Changes:
* Introduce new `populatedBy` fields - that populated by a Spring bean with the specified name.
* Implement sorting by contributors using following rule:
  * Use primary contributor's name if exists;
  * Use the first contributor's name if no primary contributor;
* Implement sorting by title.
* Add integration tests to verify sorting works as expected.
* Remove previous copy_to sorting approach for title, because for some reason it doesn't work.


src/main/resources/model/instance.json
```diff
@@ -111,16 +108,16 @@
...
+      "index": "lowercase_sort",
+      "populatedBy": "sortContributorsSetter"
+    },
+    "sort_title": {
+      "index": "lowercase_sort",
+      "populatedBy": "sortTitleSetter"
     }
   },
   "groups": { },
-  "indexMappings": {
```
pfilippovEPAM, line 117: what was wrong with this implementation?

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
```diff
@@ -114,18 +116,33 @@ private static Object getFieldValue(
...
+  private Object getPlainFieldValue(Map<String, Object> fieldData, String fieldName,
+    PlainFieldDescription desc) {
+
+    if (!desc.hasPopulatedBy()) {
+      return fieldData.get(fieldName);
+    }
+
+    final var propertySetter = fieldSetters.get(desc.getPopulatedBy());
+    if (propertySetter == null) {
+      throw new IllegalArgumentException("There is no such property setter: " + desc.getPopulatedBy());
```
pfilippovEPAM, line 138: throwing exception in terms of this type of indexing will fail the WHOLE batch, consider to refactor it

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
```diff
@@ -114,18 +116,33 @@ private static Object getFieldValue(
...
+    }
+
+    return propertySetter.getFieldValue(fieldData);
+  }
+
   @SuppressWarnings("unchecked")
-  private static Object getObjectFieldValue(
```
pfilippovEPAM, line 128: please, stop changing method signatures

src/main/java/org/folio/search/service/setter/instance/SortContributorsSetter.java
```diff
@@ -0,0 +1,43 @@
...
+      .findFirst()
+      .orElse(contributors.get(0));
+
+    return mainContributor.getName();
+  }
+
+  private List<InstanceContributors> getContributors(Map<String, Object> eventBody) {
+    final var contributors = MapUtils.getObject(eventBody, "contributors");
+
+    return ofNullable(jsonConverter.convert(contributors, InstanceContributors[].class))
```
pfilippovEPAM, line 39: ```new TypeReference<List<InstanceContributors>>{}```

src/main/resources/model/instance.json
bohdan-suprun: I'm not sure but it doesn't work as expected, you can try it yourself.

My guess that sort_title group included all the properties that has `inventorySearchType: title`.

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
bohdan-suprun: Thanks, good catch. Do you think it is fine just a log warn message?

src/main/java/org/folio/search/service/setter/instance/SortContributorsSetter.java
bohdan-suprun: Thanks, changed.

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
pfilippovEPAM: I guess that this check can be removed, it looks like as metadata validation. From my viewpoint, this should be done at the application startup or resource description loaded event. If metadata is not valid - user or developer should receive warning single time, not for each event in kafka.

bohdan-suprun: Thanks, will do so.

bohdan-suprun: Changed, the validation is part of `ResourceDescriptionService#init` logic now.

src/test/java/org/folio/search/sample/SampleInstances.java
```diff
@@ -9,7 +11,7 @@
   private static final Instance SEMANTIC_WEB = readSampleInstance("semantic_web_primer.json");
 
   public static Instance getSemanticWeb() {
-    return SEMANTIC_WEB;
+    return OBJECT_MAPPER.convertValue(SEMANTIC_WEB, Instance.class);
```
hjiebsco, line 14: Isn't `SEMANTIC_WEB` already `Instance`?

hjiebsco (approved):

src/test/java/org/folio/search/sample/SampleInstances.java
bohdan-suprun: Yes, it is. This convert needed to make a copy of the instance, so that the original instance is not changed from the outer code.

src/main/java/org/folio/search/model/metadata/PlainFieldDescription.java
```diff
@@ -57,7 +58,13 @@
    */
   private ObjectNode mappings;
 
+  private String populatedBy;
```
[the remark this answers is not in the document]
bohdan-suprun: Sure, added.

src/main/java/org/folio/search/service/converter/SearchDocumentConverter.java
```diff
@@ -59,8 +63,11 @@
   private SearchDocumentBody convert(ConversionContext context) {
     final var resourceData = context.getResourceData();
 
-    Map<String, Object> resultDocument = convertMapUsingResourceFields(resourceData,
+    Map<String, Object> baseFields = convertMapUsingResourceFields(resourceData,
       context.getResourceDescription().getFields(), context);
+    Map<String, Object> extendedFields = generateExtendedFields(context);
```
pfilippovEPAM, line 68: I would suggest to rename it to the 'searchFields', after some time it looks most correct name for these fields, from my point of view, obviously.

bohdan-suprun: renamed to `searchFields`

pfilippovEPAM (approved):

PR #21, opened 2021-02-11
MSEARCH-9: Adjust mappings for multilang fields.

bohdan-suprun:


src/main/java/org/folio/search/service/IndexService.java
```diff
@@ -56,6 +61,9 @@ public FolioCreateIndexResponse createIndex(String resourceName, String tenantId
   public FolioIndexOperationResponse updateMappings(String resourceName, String tenantId) {
     var index = getElasticsearchIndexName(resourceName, tenantId);
     var mappings = mappingHelper.getMappings(resourceName);
+
+    log.info("Updating mappings for resource [{}] for tenant [{}] - [{}]",
```
pfilippovEPAM, line 65: log format always anywhere in search should be like ```$message [$parameter: $value]```

bohdan-suprun: thanks, changed.

pfilippovEPAM (approved):

src/main/java/org/folio/search/service/es/SearchMappingsHelper.java
```diff
@@ -28,11 +29,13 @@
 public class SearchMappingsHelper {
 
   private static final String MAPPING_PROPERTIES_FIELD = "properties";
+  private static final String MULTILANG_SOURCE_FIELD = "src";
```
hjiebsco, line 32: Line 128 still uses hardcoded `src`

src/test/java/org/folio/search/controller/ConfigControllerIT.java
```diff
@@ -88,14 +90,15 @@ void shouldUseConfiguredLanguagesDuringMapping() {
...
-      .languages(languageCodes)
+      .languages(List.of("eng", "rus", "fre"))
       .title("This is title");
 
     kafkaTemplate.send(INVENTORY_INSTANCE_TOPIC, newInstance.getId(),
       eventBody(INSTANCE_RESOURCE, newInstance));
 
     final var indexedInstance = getIndexedInstanceById(newInstance.getId());
 
+    assertThat((Map<String, Object>) getMapValueByPath("title", indexedInstance), aMapWithSize(3));
```
hjiebsco, line 101: Maybe add `assertNull(getMapValueByPath("title.fre", indexedInstance));`

hjiebsco (approved):

src/main/java/org/folio/search/service/es/SearchMappingsHelper.java
bohdan-suprun: changed.

src/test/java/org/folio/search/controller/ConfigControllerIT.java
bohdan-suprun: added.

PR #25, opened 2021-02-19
Revert "FOLIO-2997 Set port for elasticsearch in default LaunchDescriptor"

ihardy:
Reverts folio-org/mod-search#24


ihardy:
@bohdan-suprun We'd like to revert this one. I think it makes sense to leave the launch descriptor w/the standard 9200 port. I had to change the standard port on the reference environments to something else because of a conflict, but most production builds should not have this issue.

bohdan-suprun (approved):

PR #26, opened 2021-02-20
MSEARCH-31 Implement indexing of items

pfilippovEPAM:
Indexing approach has been reimplemented. Now it will require additional bulk call to the inventory-storage to retrieve instance with items and holdings.


src/main/java/org/folio/search/integration/KafkaMessageListener.java
```diff
@@ -28,22 +35,61 @@
...
     indexService.indexResources(resources);
   }
 
+  @KafkaListener(
+    id = "mod-search-events-listener",
+    containerFactory = "kafkaListenerContainerFactory",
+    topics = "#{'${application.kafka.listener.events.topics}'.split(',')}",
+    groupId = "${application.kafka.listener.events.group-id}",
+    concurrency = "${application.kafka.listener.events.concurrency}")
+  public void handleEvents(List<ConsumerRecord<String, ResourceEventBody>> consumerRecords) {
```
bohdan-suprun, line 58: Are we going to remove the `instance` handler and use only this one?

src/main/java/org/folio/search/controller/FolioTenantController.java
```diff
@@ -0,0 +1,33 @@
...
+    FolioSpringLiquibase folioSpringLiquibase,
+    FolioExecutionContext context,
+    KafkaAdminService kafkaAdminService) {
+    super(folioSpringLiquibase, context);
+    this.kafkaAdminService = kafkaAdminService;
+  }
+
+  @Override
+  public ResponseEntity<String> postTenant(TenantAttributes tenantAttributes) {
+    kafkaAdminService.createKafkaTopics();
```
bohdan-suprun, line 30: Why do we need to create topics in mod-search?

src/main/resources/kafka/kafka-topics.json
```diff
@@ -0,0 +1,24 @@
...
+      "numPartitions": 50,
+      "replicationFactor": 1
+    },
+    {
+      "name": "inventory.holdings-record",
+      "numPartitions": 50,
+      "replicationFactor": 1
+    },
+    {
+      "name": "search.resource-ids",
```
bohdan-suprun, line 19: Is it reserved for future?

hjiebsco (commented):
Is this PR just for https://issues.folio.org/browse/MSEARCH-31 Item - keyword search by barcode, or the goal is to switch to id based message as documented on https://wiki.folio.org/display/DD/Search, or maybe both? Also `InventoryClient` is a placeholder, would that be covered by another ticket?

src/main/java/org/folio/search/integration/KafkaMessageListener.java
pfilippovEPAM: yes. first one is temporal solution for indexing only instances, when bulk endpoint will be implement - in term of this story I will remove previous approach.

src/main/java/org/folio/search/controller/FolioTenantController.java
pfilippovEPAM: Ideally, producer and consumer should be independent and both of them should be able to create topics using common configuration. For example, if mod-search started earlier than mod-inventory-storage - we don't want to see tons of logs with unknown_topic_or_partition errors

pfilippovEPAM:
> another

Yes,
1) After discussion with @MikhailFokanov and @bohdan-suprun, we decided to use bulk endpoint to receive must relevant instance entity with related items from inventory instead of joining this data in mod-search using elasticsearch and kafka events
2) InventoryClient for now is just a placeholder, and it will be implemented in terms of story MSEARCH-31, but a little bit later, when https://issues.folio.org/browse/MODINVSTOR-674 will be completed.

src/main/java/org/folio/search/controller/FolioTenantController.java
bohdan-suprun: Make sense, thanks.

src/main/java/org/folio/search/configuration/FeignRequestInterceptor.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class FeignRequestInterceptor implements RequestInterceptor {
+
+  private final FolioExecutionContext folioExecutionContext;
+
+  @Override
+  @SneakyThrows
+  public void apply(RequestTemplate template) {
```
bohdan-suprun, line 22: This can be removed further, after https://github.com/folio-org/folio-spring-base/pull/13.

bohdan-suprun (approved):

src/main/java/org/folio/search/service/AsyncFolioExecutionContext.java
```diff
@@ -9,7 +9,7 @@
 import org.folio.spring.FolioModuleMetadata;
 
 @RequiredArgsConstructor
-final class AsyncFolioExecutionContext implements FolioExecutionContext {
+public final class AsyncFolioExecutionContext implements FolioExecutionContext {
```
hjiebsco, line 12: Looks like there will be a same class `AsyncFolioExecutionContext.java `in the FOLIO Spring base app.

src/main/resources/model/holding.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "holding",
+  "fields": {}
```
hjiebsco, line 3: Empty fields?

src/main/resources/model/item.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "item",
+  "fields": {}
```
hjiebsco, line 3: Empty fields?

src/main/resources/swagger.api/schemas/holding.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Holding description",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Unique ID of the item record"
```
hjiebsco, line 8: `item` or `holding`?

src/main/resources/swagger.api/schemas/instance.json
```diff
@@ -217,6 +217,20 @@
...
+    "items": {
+      "type": "array",
+      "description": "List of instance items",
+      "items": {
+        "$ref": "item.json"
+      }
+    },
+    "holdings": {
+      "type": "array",
+      "description": "List of instance items",
```
hjiebsco, line 230: `items` or `holdings`?

src/main/java/org/folio/search/service/AsyncFolioExecutionContext.java
pfilippovEPAM: I've updated version to the newest 1.0.1, but I don't see it in the project, so I leave it for now, when it will be implemented, I will replace it with version from folio-spring-base

hjiebsco: That PR was closed by @bohdan-suprun this morning. :) https://github.com/folio-org/folio-spring-base/pull/13

hjiebsco (approved):

PR #29, opened 2021-02-26
MSEARCH-24 Implement search by isbn and issn identifiers

pfilippovEPAM:


hjiebsco (commented):
- The ticket [MSEARCH-24](https://issues.folio.org/browse/MSEARCH-24) mentioned four types: `ISBN, Invalid ISBN, ISSN or Invalid ISSN` but looks like this PR only added `ISBN and ISSN`, the other two identifiers have their own ids. See [InvalidIsbn](https://github.com/folio-org/mod-inventory-storage/blob/master/reference-data/identifier-types/InvalidIsbn.json) and [InvalidIssn](https://github.com/folio-org/mod-inventory-storage/blob/master/reference-data/identifier-types/InvalidIssn.json). If those are not required, please update JIRA tickets accordingly.
- For ISBN normalization, looks like hyphens should **NOT** be removed first, I made an inline suggestion, please take a look. On top of that, can you confirm with @magdazacharska if 10 digits ISBN can have qualifier as well? If yes, we need to exclude qualifier before calling the validation utility method, otherwise the code will not send the 13 digits ISBN version for indexing.

src/main/java/org/folio/search/service/setter/instance/IsbnProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+
+  private static Stream<String> normalizeIsbn(String value) {
+    if (StringUtils.isBlank(value)) {
+      return Stream.empty();
+    }
+    String cleanValue = removeHyphens(value);
+    if (isValid10DigitNumber(cleanValue)) {
+      return Stream.of(cleanValue, removeHyphens(convertTo13DigitNumber(cleanValue)));
+    }
+    return Stream.of(cleanValue);
```
hjiebsco, line 45: Removing hyphens first does not work. For example `1 86197 271-7` is a valid 10 digits ISBN but not anymore after removing hyphens. Consider below
 ```suggestion
    if (isValid10DigitNumber(value)) {
      return Stream.of(removeHyphens(value), removeHyphens(convertTo13DigitNumber(value)));
    }
    return Stream.of(removeHyphens(value));
```

Comment (pfilippovEPAM) [reply src/main/java/org/folio/search/service/setter/instance/IsbnProcessor.java]: I have reimplemented approach, now it's working with qualifiers and produce correct tokens for more isbn values (unit test data in IsbnProcessorTest has been extended too)

File: src/main/resources/elasticsearch/index/instance.json [hunk 2]
```diff
@@ -7,6 +7,12 @@
     "mapping.total_fields.limit": 250
   },
   "analysis": {
+    "char_filter": {
+      "remove_hyphen_char_filter": {
```
bohdan-suprun, line 11: Isn't we already remove the hyphens in the processor?

bohdan-suprun (approved):
Good work!

src/main/resources/elasticsearch/index/instance.json
pfilippovEPAM: it is used to remove hyphens from incoming request also

src/main/java/org/folio/search/service/setter/instance/IsbnProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+
+  private static final String SEP = "(?:[-\\s])";
+  private static final String GROUP = "(\\d{1,5})";
+  private static final String PUBLISHER = "(\\d{1,7})";
+  private static final String TITLE = "(\\d{1,6})";
+
+  private static final Pattern ISBN10_REGEX = Pattern.compile(
+    "^(?:(\\d{9}[0-9X])|(?:" + GROUP + SEP + PUBLISHER + SEP + TITLE + SEP + "([0-9X])))");
+  private static final Pattern ISBN13_REGEX = Pattern.compile(
+    "^(978|979)(?:(\\d{10})|(?:" + SEP + GROUP + SEP + PUBLISHER + SEP + TITLE + SEP + "([0-9])))");
```
hjiebsco, line 35: If these are from some web sites, can you add a link?

src/main/java/org/folio/search/service/setter/instance/IsbnProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+    resultBuilder.append(value.charAt(0));
+    for (int i = 1; i < value.length() - 1; i++) {
+      var current = value.charAt(i);
+      if (!(isBetweenNumberCharacters(value.charAt(i - 1), value.charAt(i + 1)) && isHyphenOrSpace(current))) {
+        resultBuilder.append(current);
+      }
+    }
+    resultBuilder.append(value.charAt(value.length() - 1));
+    return resultBuilder.toString();
+  }
```
hjiebsco, line 114: This method will convert `12 3` to `123` but  will not change`12  3` (note the double space in second case). Is this by design?

src/main/java/org/folio/search/service/setter/instance/IssnProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+   * @param jsonConverter {@link JsonConverter} bean
+   */
+  public IssnProcessor(JsonConverter jsonConverter) {
+    super(jsonConverter);
+  }
+
+  @Override
+  public List<String> getFieldValue(Map<String, Object> eventBody) {
+    return getInstanceIdentifiers(eventBody).stream()
+      .filter(identifier -> ISSN_IDENTIFIER_TYPE_ID.equals(identifier.getIdentifierTypeId()))
```
hjiebsco, line 33: Do you need this filter? Won't invalid issn be removed by this?

hjiebsco (commented):
Not sure if you are waiting for another round of review or not. (you did not request). Anyway I left a few inline comments.

src/main/java/org/folio/search/service/setter/instance/IssnProcessor.java
pfilippovEPAM: Fixed, forget about this part

src/main/java/org/folio/search/service/setter/instance/IsbnProcessor.java
pfilippovEPAM: I've implemented cleaning of input string from multiple spaces

hjiebsco (commented):
@pfilippovEPAM, for an invalid isbn `1 23`, the search `1 23` or `1 2*` will not be able to find it but `1*` will. Do you know the requirement for this case? The ticket https://issues.folio.org/browse/MSEARCH-24 is not clear about that. I assume this should be the regular search (#3 in the ticket) not normalized search.

hjiebsco:
> @pfilippovEPAM, for an invalid isbn `1 23`, the search `1 23` or `1 2*` will not be able to find it but `1*` will. Do you know the requirement for this case? The ticket https://issues.folio.org/browse/MSEARCH-24 is not clear about that. I assume this should be the regular search (#3 in the ticket) not normalized search.

@pfilippovEPAM, let me know what's your plan so we can move on about this PR.

src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
```diff
@@ -108,10 +110,12 @@ private QueryBuilder convertToQuery(CqlSearchRequest request, CQLNode node) {
 
   private QueryBuilder convertToTermQuery(CqlSearchRequest request, CQLTermNode node) {
     var fieldName = node.getIndex();
-    var fieldsGroup = searchFieldProvider.getFields(request.getResource(), fieldName);
+    var resource = request.getResource();
+    var fieldsGroup = searchFieldProvider.getFields(resource, fieldName);
     var fieldList = fieldsGroup.isEmpty() ? getFieldsForMultilangField(request, fieldName) : fieldsGroup;
 
-    var term = node.getTerm();
+    String term = getSearchTerm(node.getTerm(), fieldName, resource);
```
hjiebsco, line 117: Use `var` to be consistent

src/main/resources/elasticsearch/index/instance.json
```diff
@@ -18,9 +24,13 @@
         "tokenizer": "standard",
         "filter": [ "lowercase" ],
         "type": "custom"
+      },
+      "isbn_analyzer": {
+        "tokenizer": "standard",
+        "char_filter": [ "remove_hyphen_char_filter" ],
+        "filter": [ "lowercase" ]
```
hjiebsco, line 31: With the introduction of new `isbnSearchTermProcessor`, do you still need `isbn_analyzer`? It is not referenced anywhere.

hjiebsco (commented):
For `1 23`, search `1 *` does not work. Not sure how important it is though. Feel free to add a follow up user story so this one can be merged.

pfilippovEPAM:
> For 1 23, search 1 * does not work. Not sure how important it is though. Please feel free to merge this PR. I am done with reviewing it.

It is not working at other terms too, I guess we should forbid this syntax, because expected behavior of this structure is really unclear for me (it just an assumption, it should be discussed and implemented in other story)

PR #38, opened 2021-03-15
MSEARCH-55 Implement filters and facets for item and holding fields

pfilippovEPAM:
- [MSEARCH-55](https://issues.folio.org/browse/MSEARCH-55) Instance - filter/facet by item effective locations 
- [MSEARCH-56](https://issues.folio.org/browse/MSEARCH-56) Holdings - filter/facet by permanent locations 
- [MSEARCH-63](https://issues.folio.org/browse/MSEARCH-63) Item - status filters/facets

### DIFF
--- a/src/main/resources/swagger.api/schemas/item.json
+++ b/src/main/resources/swagger.api/schemas/item.json
@@ -18,6 +18,39 @@
     "effectiveLocationId": {
       "type": "string",
       "description": "Read only current home location for the item."
+    },
+    "status": {
+      "description": "The status of the item",
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "Name of the status e.g. Available, Checked out, In transit",
+          "type": "string",
+          "enum": [
+            "Aged to lost",
+            "Available",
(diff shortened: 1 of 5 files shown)


hjiebsco (approved):

src/main/resources/swagger.api/schemas/item.json
```diff
@@ -14,6 +14,43 @@
...
+      "description": "Read only current home location for the item."
+    },
+    "status": {
+      "description": "The status of the item",
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "Name of the status e.g. Available, Checked out, In transit",
+          "type": "string",
+          "enum": [
```
bohdan-suprun, line 29: I'd remove this enum restriction and use plain string instead, the reason is that from time to time new statuses are added and if we don't sync them up we will get unmarshalling error.

bohdan-suprun: Since Pavel is on vacation, I've decided to remove the enum constraint. We can revert it if we decide it required.

PR #40, opened 2021-03-16
MSEARCH-76: Fetch instance ISBN/ISSN identifiers by name.

bohdan-suprun:
ISBN/ISSN identifiers ids are different per environments. This PR removes hardcoded identifier ids and instead queries identifiers by their name and then uses the correct IDs.


hjiebsco (approved):
Curious, do all changes in this PR are related to isbn/issn? I mean the refactoring work related to converter. 
BTW those different ids from Bugfest could be a legacy issue. They were likely created at really early days before the later formalized reference data (preferred).

bohdan-suprun:
@hjiebsco,

> Curious, do all changes in this PR are related to isbn/issn? 

Well, yes, they related, we have to begin folio scope in order to call the API in the ISBN processor, that's mostly the goal of the changes in processor. 

> BTW those different ids from Bugfest could be a legacy issue. They were likely created at really early days before the later formalized reference data (preferred).

Do you think we should rather fix data instead of the code?

hjiebsco:
> Do you think we should rather fix data instead of the code?

No. If we can count on those identifier names to be constant/fixed values, names are more flexible. BTW I noticed you changed the test sample file `samples/semantic-web-primer/instance.json` to match Bugfest. Maybe that is not necessary.

bohdan-suprun:
> BTW I noticed you changed the test sample file samples/semantic-web-primer/instance.json to match Bugfest. Maybe that is not necessary.

Agree, reverted.

PR #43, opened 2021-03-17
MSEARCH-34: Add holdings.discoverySuppress filter and facet property.

bohdan-suprun:
## Changes:

* Add facet and filter for `holdings.discoverySuppress` property;
* Add `default` element to mappings that allow to set default value for a property when no property or null.


src/main/java/org/folio/search/utils/CollectionUtils.java
```diff
@@ -24,4 +24,17 @@
...
+   *
+   * @param map - The map.
+   * @param key - Key to check.
+   * @param <K> - Key type
+   * @param <V> - Value type
+   * @return true if key has a non-null value associated, otherwise false.
+   */
+  public static <K, V> boolean hasNoValue(Map<K, V> map, K key) {
+    return map.get(key) == null;
+  }
```
hjiebsco, line 39: Not sure if a simple method like this is necessary.

src/main/resources/swagger.api/schemas/holding.json
```diff
@@ -10,6 +10,10 @@
     "permanentLocationId": {
       "type": "string",
       "description": "The permanent shelving location in which an item resides."
+    },
+    "discoverySuppress": {
+      "type": "boolean",
+      "description": "records the fact that the record should not be displayed in a discovery system"
```
hjiebsco, line 16: Maybe change `records the fact` to `Indicates`?

hjiebsco (approved):

src/main/java/org/folio/search/utils/CollectionUtils.java
bohdan-suprun: Removed and replaced with commons-MapUtils call.

src/main/resources/swagger.api/schemas/holding.json
bohdan-suprun: Changed.

PR #54, opened 2021-03-30
MSEARCH-81: Implement holdings.fullCallNumber search option.

bohdan-suprun:


hjiebsco (approved):

src/main/resources/model/instance.json
```diff
@@ -257,6 +257,15 @@
               "index": "multilang"
             }
           }
+        },
+        "callNumberPrefix": {
+          "index": "source"
```
pfilippovEPAM, line 262: Why these values included in source? From my viewpoint, they won't be returned to UI in general use

pfilippovEPAM (approved):
Overall, looks good to me, except one comment with redundant values in source document

src/main/resources/model/instance.json
bohdan-suprun: Removed them, we can add them again when needed.

PR #56, opened 2021-04-01
MSEARCH-72: Throw more explicit error if tenant is not initialized yet

bohdan-suprun:
When kafka listener starts consuming events from the topic but tenant is not initialized yet we throw a `TenantNotInitializedException` vs a sql exception which confuses folks.

### DIFF
--- a/src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
+++ b/src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
@@ -29,13 +29,13 @@ public class KafkaErrorHandler implements KafkaListenerErrorHandler {
     // which is treated as a grammar exception.
     // It is possible to have false positives here, when an invalid SQL is used
     // but anyway it results in retrying the consume operation
-    return getThrowableList(exception).stream().anyMatch(ex -> ex instanceof SQLGrammarException);
+    return getThrowableList(exception).stream().anyMatch(SQLGrammarException.class::isInstance);
   }
 
   @SuppressWarnings("unchecked")
   private String[] getTenantNames(Message<?> message) {
     if (!isSupportedType(message)) {
-      return null;
+      return new String[0];
     }
     var consumerRecords = (List<ConsumerRecord<String, ResourceEventBody>>) message.getPayload();
     return consumerRecords.stream()
@@ -52,7 +52,7 @@ public class KafkaErrorHandler implements KafkaListenerErrorHandler {
     }
 
     var payloads = (List) message.getPayload();
-    return payloads.size() > 0 && payloads.get(0) instanceof ConsumerRecord
+    return !payloads.isEmpty() && payloads.get(0) instanceof ConsumerRecord
       && ((ConsumerRecord) payloads.get(0)).value() instanceof ResourceEventBody;
   }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.messaging.Message;
+import org.springframework.stereotype.Component;
+
+@Component
+public class KafkaErrorHandler implements KafkaListenerErrorHandler {
+  @Override
+  public Object handleError(Message<?> message, ListenerExecutionFailedException exception) {
+    if (isTenantNotInitializedYet(exception)) {
+      throw new TenantNotInitializedException(getTenantNames(message));
+    } else {
```
pfilippovEPAM, line 21: else condition can be removed

pfilippovEPAM (approved):

src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
```diff
@@ -0,0 +1,58 @@
...
+    // which is treated as a grammar exception.
+    // It is possible to have false positives here, when an invalid SQL is used
+    // but anyway it results in retrying the consume operation
+    return getThrowableList(exception).stream().anyMatch(SQLGrammarException.class::isInstance);
+  }
+
+  @SuppressWarnings("unchecked")
+  private String[] getTenantNames(Message<?> message) {
+    if (!isSupportedType(message)) {
+      return new String[0];
```
hjiebsco, line 38: Curious, in this case, Is it still a `TenantNotInitializedException` even though there are no tenant names?

src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
```diff
@@ -0,0 +1,58 @@
...
+      throw exception;
+    }
+  }
+
+  private boolean isTenantNotInitializedYet(ListenerExecutionFailedException exception) {
+    // In case when the schema for tenant is not created
+    // DB will throw an error saying that a table does not exists
+    // which is treated as a grammar exception.
+    // It is possible to have false positives here, when an invalid SQL is used
+    // but anyway it results in retrying the consume operation
```
hjiebsco, line 31: Would those potential false positives make the logs more confusing?

hjiebsco (approved):

src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
bohdan-suprun: Make sense, removed this part.

src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
bohdan-suprun: I'm now specifying the original cause, so we can debug it when needed.

src/main/java/org/folio/search/integration/error/KafkaErrorHandler.java
bohdan-suprun: changed

PR #58, opened 2021-04-01
MSEARCH-85 cannot read mod-search.csv from classpath

hjiebsco:
This PR is to address below error in Bugfest env when enabling mod-search for tenant. See https://issues.folio.org/browse/MSEARCH-85
`Caused by: java.io.FileNotFoundException: class path resource [permissions/mod-search.csv] cannot be resolved to absolute file path because it does not reside in the file system: jar:file:/usr/ms/ms.jar!/BOOT-INF/classes!/permissions/mod-search.csv`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ target/
 !.mvn/wrapper/maven-wrapper.jar
 !**/src/main/**/target/
 !**/src/test/**/target/
+bin/
+.checkstyle
 
 ### STS ###
 .apt_generated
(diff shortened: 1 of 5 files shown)


bohdan-suprun (approved):

hjiebsco:
@bohdan-suprun, do we need a 1.1.1 patch release for this, or it can be rolled into 1.2.0 release?

bohdan-suprun:
> @bohdan-suprun, do we need a 1.1.1 patch release for this, or it can be rolled into 1.2.0 release?

@hjiebsco, I'll prepare 1.2.0 today.

pfilippovEPAM (approved):

PR #61, opened 2021-04-06
MSEARHC-45: Implement DELETE/GET /_/tenant endpoint

bohdan-suprun:
## Changes:
* Expose `DELETE/GET` `/_/tenant` endpoints;
* Remove elasticsearch index on `DELETE` `/_/tenant`;
* Use Posrgres testcontainer instead of H2 because base DELETE tenant uses Postgres specific features;
* Move postgres/kafka/elasticsearch container management to extensions, so that they can be reused in other component tests.


src/test/java/org/folio/search/service/IndexServiceTest.java
```diff
@@ -111,4 +111,24 @@ void reindexInventory_positive() {
 
     verify(instanceStorageClient).submitReindex();
   }
+
+  @Test
+  void shouldDropIndexWhenExists() {
+    var index = "instance_diku";
```
pfilippovEPAM, line 117: maybe use existing constant INDEX_NAME from TestConstants class?

src/test/java/org/folio/search/support/base/BaseIntegrationTest.java
```diff
@@ -40,62 +35,36 @@
...
 import org.springframework.test.web.servlet.ResultActions;
-import org.testcontainers.containers.GenericContainer;
-import org.testcontainers.containers.KafkaContainer;
-import org.testcontainers.images.builder.ImageFromDockerfile;
-import org.testcontainers.utility.DockerImageName;
 
 @Log4j2
 @SpringBootTest
 @AutoConfigureMockMvc
+@EnablePostgres
```
pfilippovEPAM, line 45: nice solution :thumbsup:

pfilippovEPAM (approved):

hjiebsco (approved):

hjiebsco:
Noticed that tenant API is still at 1.2 not 2.0 yet.

bohdan-suprun:
> Noticed that tenant API is still at 1.2 not 2.0 yet.

Yes, the folio-spring-base implements only this version.

I thought the 1.2 is still supported with 2.0, right?

hjiebsco:
> > Noticed that tenant API is still at 1.2 not 2.0 yet.
> 
> Yes, the folio-spring-base implements only this version.
> 
> I thought the 1.2 is still supported with 2.0, right?

That's right @bohdan-suprun. I was just wondering since other modules have moved to 2.0.

bohdan-suprun:
> That's right @bohdan-suprun. I was just wondering since other modules have moved to 2.0.

I think most of non-RMB modules are still with 1.2 version. Probably we need a story for folio-spring-base to implement it for all of the modules. What do you think?

hjiebsco:
> > That's right @bohdan-suprun. I was just wondering since other modules have moved to 2.0.
> 
> I think most of non-RMB modules are still with 1.2 version. Probably we need a story for folio-spring-base to implement it for all of the modules. What do you think?

Makes sense. Keep modules consistent would be nice.

bohdan-suprun:
@hjiebsco, https://issues.folio.org/browse/FOLSPRINGB-17 created.

src/test/java/org/folio/search/service/IndexServiceTest.java
bohdan-suprun: Thanks, changed.

PR #69, opened 2021-04-09
MSEARCH-97: Allow passing OKAPI_URL to login system user 

bohdan-suprun:
## Changes
* Convert existing `SystemUserService` to support in-memory store of user credentials;
* Require `OKAPI_URL` env variable;
* Remove `system_user` table;
* Remove token caches and use Spring caching capabilities;
* Move common logic related to user create/update and login to separate `PrepareSystemUserService` class.


hjiebsco (approved):
If I understand correctly, the logic implemented is 1) if there is no env OkapiUrl specified, the db one will be used. That means it won't self correct the stale Okapi url stored in db since it is only populated during tenant init, correct? 2) if env OkapiUrl is used, the in memory one will be used, and it won't correct the stale url in db either, correct? Overall, I think the intention of supporting both db and in memory in a smooth transition way is really good. It would also be nice if we just switch to in-memory in one shot to keep the logic simple and clean. With that said, I am fine if you want to merge as is.

bohdan-suprun:
> 1) if there is no env OkapiUrl specified, the db one will be used. That means it won't self correct the stale Okapi url stored in db since it is only populated during tenant init, correct? 

Yes correct. We can not correct the state until an instance receive any request from user that comes from OKAPI. That even can not happen at all.

2) if env OkapiUrl is used, the in memory one will be used, and it won't correct the stale url in db either, correct? 

Yes, because it does not make sense at all, I think since this mode is disabled.

> Overall, I think the intention of supporting both db and in memory in a smooth transition way is really good. It would also be nice if we just switch to in-memory in one shot to keep the logic simple and clean. 

I'm fine with this approach if it waits. I'll prioritize https://issues.folio.org/browse/FOLIO-3059. (I'd appreciate if you help me with pushing this)

hjiebsco:
> Yes, because it does not make sense at all, I think since this mode is disabled.

True but this will leave some stale data in db, and if later runs does not specify the OkapiUrl env var, it gets messy.

> I'm fine with this approach if it waits. I'll prioritize https://issues.folio.org/browse/FOLIO-3059. (I'd appreciate if you help me with pushing this)

I voted and commented to that ticket, but I think this is where the scrum master and PO will shine.

pfilippovEPAM (approved):

bohdan-suprun:
As discussed I've removed bi-support of DB and non-DB mode and now we support only non-DB mode.

Since mod-search is disabled for reference builds (https://issues.folio.org/browse/MSEARCH-86) we're safe to merge the PR I think, what do you think @hjiebsco?

hjiebsco (approved):

hjiebsco:
> As discussed I've removed bi-support of DB and non-DB mode and now we support only non-DB mode.
> 
> Since mod-search is disabled for reference builds (https://issues.folio.org/browse/MSEARCH-86) we're safe to merge the PR I think, what do you think @hjiebsco?

Sounds good to me @bohdan-suprun. Thanks for the PR change.

PR #70, opened 2021-04-09
MSEARCH-86: Remove not required env variables from module descriptor.

bohdan-suprun:
> Ok, if my assumptions are correct and deployment tool copies env variables from module descriptor as they are, than the quickest fix is to leave only those variables that will be overridden by OKAPI: DB connection properties and remove all other properties.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -216,26 +216,6 @@
         "value": "9092"
       },
       {
-        "name": "ELASTICSEARCH_HOST",
-        "value": "elasticsearch"
-      },
-      {
-        "name": "ELASTICSEARCH_PORT",
-        "value": "9200"
-      },
-      {
-        "name": "ELASTICSEARCH_URL",
-        "value": "http://elasticsearch:9200"
-      },
-      {
-        "name": "ELASTICSEARCH_USERNAME",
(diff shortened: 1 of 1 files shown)


hjiebsco (commented):
I do not know the reason to remove these env vars, but when we set up a new module, we usually check the module descriptor to see what's new and add new ones to our deployment configuration. Alternatively, if every module documents the required env vars on README properly, we can use that.

bohdan-suprun:
> I do not know the reason to remove these env vars, but when we set up a new module, we usually check the module descriptor to see what's new and add new ones to our deployment configuration. Alternatively, if every module documents the required env vars on README properly, we can use that.

That's my expectation as well. However, the deployment process seems get the env variables from the descriptor and propagates them to the container as they are. It causes issues in code, e.g. ELASTICSEARCH_URL takes precision over ELACTICSEARCH_HOST/PORT, so that the  hardcoded sample value for ELASTICSEARCH_URL was used.

pfilippovEPAM (approved):

hjiebsco (commented):
It is a bit inconsistent to keep DB and Kafka but remove ES. But I will leave devops team to decide.

bohdan-suprun:
Closing as @funkymalc working on fix from their side.

PR #72, opened 2021-04-15
MSEARCH-93 Implement retrieving instance ids as json stream

pfilippovEPAM:


src/main/java/org/folio/search/repository/SearchRepository.java
```diff
@@ -36,8 +51,53 @@ public SearchResponse search(ResourceRequest resourceRequest, SearchSourceBuilde
...
+    consumer.accept(getInstanceIds(searchHits));
+
+    while (isNotEmpty(searchHits)) {
+      var scrollRequest = new SearchScrollRequest(scrollId).scroll(KEEP_ALIVE_INTERVAL);
+      var scrollResponse = performExceptionalOperation(
+        () -> elasticsearchClient.scroll(scrollRequest, DEFAULT), index, "scrollApi");
+      scrollId = scrollResponse.getScrollId();
+      searchHits = scrollResponse.getHits().getHits();
+      consumer.accept(getInstanceIds(searchHits));
+    }
```
bohdan-suprun, line 85: Can we use do/while here to remove duplicated logic before the while?

src/main/java/org/folio/search/utils/SearchUtils.java
```diff
@@ -18,6 +18,8 @@
   public static final String PLAIN_MULTILANG_PREFIX = "plain_";
   public static final String DOT = ".";
 
+  public static final int MAX_ELASTICSEARCH_QUERY_SIZE = 10_000;
```
bohdan-suprun, line 21: Is it a limit? If yes, I think we should confirm it with Magda. As far as I remember for inventory equivalent API we used 65 000 as a limit.

descriptors/ModuleDescriptor-template.json
```diff
@@ -61,6 +61,13 @@
           ],
           "pathPattern": "/search/instances/facets",
           "permissionsRequired": [ "search.instances.facets.collection.get" ]
+        },
+        {
+          "methods": [
+            "GET"
+          ],
+          "pathPattern": "/search/instances/ids",
+          "permissionsRequired": [ "search.instances.ids.collections.get" ]
```
hjiebsco, line 70: Do you mean `collections` or just `collection`? Other similar perms used the latter.

src/main/java/org/folio/search/repository/SearchRepository.java
hjiebsco, line 85: Would it work if replace both `consumer.accept(getInstanceIds(searchHits));` with just one right after line 78?

src/main/resources/swagger.api/examples/ResourceIdsResult.sample
```diff
@@ -0,0 +1,14 @@
...
+    },
+    {
+      "id": "e60063f6-db50-4a2a-93f6-6b27ab0420f2"
+    },
+    {
+      "id": "5fab345a-30a4-4aee-8667-1e33d4627ce9"
+    }
+  ],
+  "totalRecords": 3
+}
```
hjiebsco, line 14: Would an array of String work? The extra "id" seems a bit redundant, or maybe you just want to be flexible?

hjiebsco (approved):

src/main/java/org/folio/search/utils/SearchUtils.java
pfilippovEPAM: yes, this is limit from elasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/7.12/paginate-search-results.html

src/main/resources/swagger.api/examples/ResourceIdsResult.sample
pfilippovEPAM: The same contract is used for request, so I've decided to keep it for UI to make it easier to integrate: 

 `.../record-bulk/ids?limit=2147483647&query=(keyword all "book") sortby title"`

src/main/java/org/folio/search/repository/SearchRepository.java
pfilippovEPAM: alternate solution would be like:

```
  /**
   * Executes scroll request to elasticsearch and transforms it to the list of instance ids.
   *
   * @param request resource request as {@link ResourceRequest} object.
   * @param source elasticsearch search source as {@link SearchSourceBuilder} object.
   */
  public void streamResourceIds(ResourceRequest request, SearchSourceBuilder source, Consumer<List<String>> consumer) {
    var index = getElasticsearchIndexName(request);
    var searchRequest = new SearchRequest()
        .scroll(new Scroll(KEEP_ALIVE_INTERVAL))
        .routing(request.getTenantId())
        .source(source)
        .indices(index);
    String scrollId = null;
    SearchHit[] searchHits;
    do {
      var searchResponse = executeSearchRequest(searchRequest, index, scrollId);
      searchHits = searchResponse.getHits().getHits();
      scrollId = searchResponse.getScrollId();
      consumer.accept(getInstanceIds(searchHits));
    } while (isNotEmpty(searchHits));
    clearScrollAfterStreaming(index, scrollId);
  }

  private SearchResponse executeSearchRequest(SearchRequest searchRequest, String index, String scrollId) {
    if (scrollId == null) {
      return performExceptionalOperation(
          () -> elasticsearchClient.search(searchRequest, DEFAULT), index, "searchApi")
    }
    var scrollRequest = new SearchScrollRequest(scrollId).scroll(KEEP_ALIVE_INTERVAL);
    return performExceptionalOperation(
        () -> elasticsearchClient.scroll(scrollRequest, DEFAULT), index, "scrollApi");
  }
```

From my viewpoint, it's more complicated to read that using suggested

bohdan-suprun (approved):

PR #74, opened 2021-04-20
MSEARCH-86: Increase retry period.

bohdan-suprun:
Increase retry period to 20 seconds

### DIFF
--- a/src/main/java/org/folio/search/configuration/KafkaConfiguration.java
+++ b/src/main/java/org/folio/search/configuration/KafkaConfiguration.java
@@ -2,7 +2,6 @@ package org.folio.search.configuration;
 
 import static org.apache.kafka.clients.consumer.ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG;
 import static org.apache.kafka.clients.consumer.ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG;
-import static org.springframework.util.backoff.FixedBackOff.UNLIMITED_ATTEMPTS;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


hjiebsco (approved):
Just be sure, can you verify this handles gracefully if someone manually deletes/expires messages using Kafka tools?

pfilippovEPAM (approved):

src/main/java/org/folio/search/configuration/KafkaConfiguration.java
```diff
@@ -43,9 +47,17 @@
...
     factory.setBatchListener(true);
     factory.setConsumerFactory(jsonNodeConsumerFactory());
+    factory.setBatchErrorHandler(kafkaFactoryErrorHandler());
     return factory;
   }
 
+  private BatchErrorHandler kafkaFactoryErrorHandler() {
+    var errorHandler = new SeekToCurrentBatchErrorHandler();
+    // infinity retry every 20 seconds
+    errorHandler.setBackOff(new FixedBackOff(20_000, UNLIMITED_ATTEMPTS));
```
pfilippovEPAM, line 57: everything looks good, except this line, maybe we will extract this property to application.yml?

bohdan-suprun: Make sense, extracted.

bohdan-suprun:
> Just be sure, can you verify this handles gracefully if someone manually deletes/expires messages using Kafka tools?

You're right, It seems I misunderstood documentation. If we set UNLIMMITTED renewals than producer will keep records in memory and redeliver without syncing metadata with Kafka broker.

I have set the attempts count to 10 so I believe it should retry 10 times and then update offset for topic so that the records will be delivered again but within another attempt (when pool() is called next time).

I'll double check it on Rancher tomorrow.

bohdan-suprun:
Verified, works as expected on Rancher

PR #75, opened 2021-04-21
MSEARCH-112: Check that ES index exists before any indexing logic.

bohdan-suprun:
When Indexing resources make sure ES index exists before starting any logic: converting resources, calling inventory, etc.


hjiebsco (approved):
Consider to make a constant and reuse that long error message `Cancelling bulk operation [reason: Cannot index resources for non existing indices, tenant not initialized? [indices=%s]]`

pfilippovEPAM (approved):

PR #80, opened 2021-04-28
MSEARCH-67: Add ASCII folding filter.

bohdan-suprun:
## Changes:
* Add ASCII folding filter for multilang fields (src property);
* `Src` field is always present for multilang now.


pfilippovEPAM (commented):
Everything looks good, but I think it is possible to add test that ascii folding filter is working

hjiebsco (approved):
LGTM

pfilippovEPAM (approved):

pfilippovEPAM:
> > Everything looks good, but I think it is possible to add test to verify that ascii folding filter is working
> 
> @pfilippovEPAM, there is an integration test, see src/test/java/org/folio/search/controller/SearchInstanceIT.java, or do you mean an another test?

Oh, I see now, approved.

PR #87, opened 2021-05-13
MSEARCH-127: Describe on-premise ES configuration in readme.

bohdan-suprun:
Add readme about on-premise ES config and recommended production configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,13 @@ request parameters:
 | query         | Yes                  | A CQL query to execute                                                          |
 | limit         | No (default to 100)  | Maximum number of records to fetch                                              |
 | offset        | No (default to 0)    | Instructs to skip first N records that matches the query                        |
-| expandAll     | No (default to false)| If false than only basic instance properties returned, otherwise all properties |
+| expandAll     | No (default to false)| If false than only _*basic_ instance properties returned, otherwise all properties |
+
+> *_Basic fields are following:_
+> * _id_
+> * _title_
+> * _contributors_
+> * _publication_
 
 We use CQL query for search queries, see [documentation](https://github.com/folio-org/raml-module-builder#cql-contextual-query-language)
 for more details.
--- a/README.md
+++ b/README.md
@@ -57,6 +57,48 @@ These languages will be added on tenant init and applied to index. Example usage
 
 ## Deploying
 
+## Configuring Elasticsearch
+
+### Configuring on-premise Elasticsearch instance
+
+It is required to install some required plugins for your ES instance, here is the list:
+* analysis-icu
+* analysis-kuromoji
+* analysis-smartcn
+* analysis-nori
+* analysis-phonetic
+
+You can find sample Dockerfile in `docker/elasticsearch/Dockerfile` or install plugins manually:
+```bash
+${ES_HOME}/bin/elasticsearch-plugin install --batch \
+  analysis-icu \
+  analysis-kuromoji \
+  analysis-smartcn \
+  analysis-nori \
+  analysis-phonetic
+```
+
+See also [Install Elasticsearch with Docker](https://www.elastic.co/guide/en/elasticsearch/reference/7.5/docker.html).
+
+There is an alternative ES image from Bitname - [bitnami/elasticsearch](https://hub.docker.com/r/bitnami/elasticsearch),
+that does not require extending dockerfile but has an env variable `ELASTICSEARCH_PLUGINS` to specify plugins to install.
+
+For production installations it is strongly recommended enabling security for instance and set-up user/password. The user
+must have at least following permissions:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,6 +57,48 @@ These languages will be added on tenant init and applied to index. Example usage
...
+
+For production installations it is strongly recommended enabling security for instance and set-up user/password. The user
+must have at least following permissions:
+* Create/delete/update index and mappings for it.
+* Create/update/delete documents in index.
+
+### Recommended production set-up
+
+The data nodes Elasticsearch configuration completely depends on the data.
+If there are 7 mln of instances the configuration with 2 nodes with 8Gb RAM and 500 Gb disk (AWS m5.large) works well.
```
hjiebsco, line 94: Does the 2 data nodes work with production setup (dedicated master node)? The default index config has `"number_of_replicas": 2`. See https://github.com/folio-org/mod-search/blob/master/src/main/resources/elasticsearch/index/instance.json#L4.

bohdan-suprun: I quoted @MikhailFokanov's reply to John in slack, @MikhailFokanov could you clarify?

predictron-cloud: @hjiebsco, @bohdan-suprun I wrote that 2 nodes, which both are master and data is the most basic configuration. Also I wrote, that for fault tolerance Elastic needs dedicated master nodes not to have split brain. For full fault tolerance there should be 3 data nodes to have ability to write and read to index even if one node is lost. But maybe guys should consider tradeoff between full fault tolerance and budget for Elastic AWS hosting.

My comment was:
|The data nodes Elasticsearch configuration completely depends on the data. If there are 7 mln of instances the configuration with 2 nodes with 8Gb RAM and 500 Gb disk (AWS m5.large) works well. We performed performance tests for this configuration and it showed good results. But I would recommend to performing additional performance testing (try to reindex and search with different configurations) with different type of nodes, and see what configuration is sufficient for what data volume. **Also for fault tolerance Elastic needs dedicated master nodes (not to have quorum problem which is called split brain) with less powerful configuration (see https://www.elastic.co/guide/en/cloud-enterprise/current/ece-ha.html).**
This described suggestion **the most basic configuration**, but having more expensive configuration could be money wasting if the basic one works well for the data.

MikhailFokanov: @hjiebsco, @bohdan-suprun I wrote that 2 nodes, which both are master and data is the most basic configuration. Also I wrote, that for fault tolerance Elastic needs dedicated master nodes not to have split brain. For full fault tolerance there should be 3 data nodes to have ability to write and read to index even if one node is lost. But maybe guys should consider tradeoff between full fault tolerance and budget for Elastic AWS hosting.

My comment was:
The data nodes Elasticsearch configuration completely depends on the data. If there are 7 mln of instances the configuration with 2 nodes with 8Gb RAM and 500 Gb disk (AWS m5.large) works well. We performed performance tests for this configuration and it showed good results. But I would recommend to performing additional performance testing (try to reindex and search with different configurations) with different type of nodes, and see what configuration is sufficient for what data volume. Also for fault tolerance Elastic needs dedicated master nodes (not to have quorum problem which is called split brain) with less powerful configuration (see https://www.elastic.co/guide/en/cloud-enterprise/current/ece-ha.html).
This described suggestion the most basic configuration, but having more expensive configuration could be money wasting if the basic one works well for the data.

hjiebsco: > @hjiebsco, @bohdan-suprun I wrote that 2 nodes, which both are master and data is the most basic configuration. Also I wrote, that for fault tolerance Elastic needs dedicated master nodes not to have split brain. For full fault tolerance there should be 3 data nodes to have ability to write and read to index even if one node is lost. But maybe guys should consider tradeoff between full fault tolerance and budget for Elastic AWS hosting.
> 
> My comment was:
> The data nodes Elasticsearch configuration completely depends on the data. If there are 7 mln of instances the configuration with 2 nodes with 8Gb RAM and 500 Gb disk (AWS m5.large) works well. We performed performance tests for this configuration and it showed good results. But I would recommend to performing additional performance testing (try to reindex and search with different configurations) with different type of nodes, and see what configuration is sufficient for what data volume. Also for fault tolerance Elastic needs dedicated master nodes (not to have quorum problem which is called split brain) with less powerful configuration (see https://www.elastic.co/guide/en/cloud-enterprise/current/ece-ha.html).
> This described suggestion the most basic configuration, but having more expensive configuration could be money wasting if the basic one works well for the data.

Consider to mention it explicitly that the 2 nodes performance test was done in a mixed master/data nodes mode.

hjiebsco (approved):

pfilippovEPAM (approved):
Looks good to me.

bohdan-suprun:
@hjiebsco, thanks. I've mentioned about the mixed mode.

PR #88, opened 2021-05-14
MSEARCH-125 Retrieve facets values for selected facets

pfilippovEPAM:
### Purpose
If facet from the second page is selected, its value is not returned for the first request.

### Approach
- Adjust FacetQueryBuilder.java to create 2 facet aggregations (with included terms and without them)
- Adjust ElasticsearchFacetConverter.java to parse received aggregation results
- Add integration test case to verify that feature working
- Cover new functionality with unit tests


src/main/java/org/folio/search/model/Pair.java
```diff
@@ -16,10 +16,10 @@
   /**
    * Left value of pair object.
    */
```
bohdan-suprun, line 18: Do we need to update comments as well?

src/main/java/org/folio/search/service/converter/ElasticsearchFacetConverter.java
```diff
@@ -76,4 +83,23 @@ private static FacetItem facetItem(Bucket bucket) {
   private static FacetResult facetResult(Map<String, Facet> facets) {
     return new FacetResult().facets(facets).totalRecords(facets.size());
   }
+
+  private static Map<String, Facet> mergeSelectedAndNormalFacets(Map<String, Facet> facets) {
+    var result = new LinkedHashMap<String, Facet>();
+
+    facets.forEach((key, value) -> {
+      var isSelectedAggregation = isSelectedTermsAggregation(key);
+      var finalKey = isSelectedAggregation ? key.substring(SELECTED_AGG_PREFIX.length()) : key;
```
bohdan-suprun, line 92: Are we adding this synthetic prefix to distinguish selected vs all facets?

src/main/java/org/folio/search/service/converter/FacetQueryBuilder.java
```diff
@@ -46,16 +57,18 @@
...
+  private List<AggregationBuilder> getFacetAggregation(CqlFacetServiceRequest request,
     QueryBuilder query, Pair<String, Integer> facetAndLimit) {
-    var field = facetAndLimit.getLeft();
+    var field = facetAndLimit.getFirst();
     validateFacetField(field, request.getResource());
-    return getFilterQueryForFacet(field, query)
-      .map(filter -> filterAggregation(filter, facetAndLimit))
-      .orElse(termsAggregation(field, facetAndLimit));
+    var filterAndFacetTerms = getFilterQueryAndFacetTerms(field, query);
+    return filterAndFacetTerms.getFirst() != null
```
bohdan-suprun, line 69: What does it mean when Facet name is null? Should we add a comment?

src/main/java/org/folio/search/service/converter/ElasticsearchFacetConverter.java
```diff
@@ -76,4 +83,23 @@ private static FacetItem facetItem(Bucket bucket) {
...
+      var facet = result.computeIfAbsent(finalKey, v -> facet(new ArrayList<>()));
+      var startIndex = isSelectedAggregation ? 0 : facet.getValues().size();
+      facet.getValues().addAll(startIndex, value.getValues());
+      facet.setTotalRecords(facet.getTotalRecords() + value.getValues().size());
```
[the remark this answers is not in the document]
pfilippovEPAM: I am not a really big fan of comments in code, but I guess I can left it in the javadoc in the implementation details

src/main/java/org/folio/search/service/converter/FacetQueryBuilder.java
pfilippovEPAM: It is not facet name, first value - is the filter query for facet, it can be null if user don't specify filer conditions

src/main/java/org/folio/search/service/converter/ElasticsearchFacetConverter.java
pfilippovEPAM: yes, it is common practice to segregate aggregation values (also, Elasticsearch doesn't allow to user to have similar aggregation names at the same level)

bohdan-suprun (approved):

hjiebsco (approved):
LGTM

PR #89, opened 2021-05-18
MSEARCH-122: Introduce field type references for index mappings file. 

bohdan-suprun:
## Approach:

Introduce field type references for index mappings (resource description). The idea is to add a new section - `fieldTypes` - in the json file and then referencing it by name via `$type` property.

Example:
```javascript
"fieldTypes" : {
   "typeName1": {
      // I.e. a plain field
      "searchTypes": [ "facet", "filter" ],
      "index": "bool",
      "default": false
   },
   // i.e. an object field
   "typeName2": {
      "type": "object",
      "properties": {
        "property1": {
          "searchTypes": "filter",
          "index": "date"
        }
   },
  "typeNameN": {}
}
```

As it can be seen both plain and object fields are supported.

The type can be than used in a `field`, e.g.: 
```javascript
"fields": {
    "field1": {
       "$type": "typeName1"
    },
    "field2": {
       "$type": "typeName2"
    },
    "field1_2": {
       "$type": "typeName1"
    }
}
``` 

## Changes:
1. Add `fieldType` to base `FieldDescription.java` class.
2. Add `fieldTypes` to `ResourceDescription` class.
3. Implement `PostProcessResourceDescriptionConverter` that does all the linkage logic.
4. Use the feature for `discoverySuppress` and `staffSuppress` properties.
5. Implement tests.


src/main/java/org/folio/search/model/metadata/FieldDescription.java
```diff
@@ -24,4 +28,12 @@
    * Field type.
    */
   private FieldType type;
+
+  /**
+   * Pre-defined type for the field, if applicable.
+   */
+  @JsonProperty("$type")
+  @Getter(AccessLevel.PACKAGE)
+  @Setter(AccessLevel.PACKAGE)
```
pfilippovEPAM, line 37: I think that using package level setters and getters in this model is not necessary

src/main/java/org/folio/search/model/metadata/ResourceDescription.java
```diff
@@ -57,35 +55,12 @@
...
   private Map<String, JsonNode> indexMappings;
 
+  /**
+   * Defined reusable types that can be referenced via $type property.
+   */
+  private Map<String, FieldDescription> fieldTypes = Collections.emptyMap();
+
   @JsonIgnore
-  @Setter(value = AccessLevel.NONE)
+  @Setter(AccessLevel.PACKAGE)
```
pfilippovEPAM, line 64: Why do we need package level setter here, why it cannot be public?

src/main/java/org/folio/search/model/metadata/PostProcessResourceDescriptionConverter.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private void resolveFieldByType(ResourceDescription desc, Map<String, FieldDescription> fields) {
+    for (var entry : fields.entrySet()) {
+      if (isNotBlank(entry.getValue().getFieldType())) {
+        entry.setValue(getFieldByType(desc, entry.getValue().getFieldType()));
+      }
+
+      // Resolve nested properties recursively
+      if (entry.getValue() instanceof ObjectFieldDescription) {
+        resolveFieldByType(desc, ((ObjectFieldDescription) entry.getValue()).getProperties());
```
pfilippovEPAM, line 34: `entry.getValue()` is used 4 times, I guess it is worth to extract it to variable

src/main/java/org/folio/search/model/metadata/ResourceDescription.java
bohdan-suprun: It is an calculated field that depends on `fields`, I would not allow it to be modified.

src/main/java/org/folio/search/model/metadata/PostProcessResourceDescriptionConverter.java
bohdan-suprun: Fixed

src/main/java/org/folio/search/model/metadata/FieldDescription.java
bohdan-suprun: Modified to be public

src/main/java/org/folio/search/model/metadata/ResourceDescription.java
pfilippovEPAM: if so, flatten fields should be collected to unmodifiable map at least to prevent further modifications

pfilippovEPAM (approved):

src/main/java/org/folio/search/model/metadata/ResourceDescription.java
bohdan-suprun: Ah, yes, totally makes sense.

hjiebsco (approved):
LGTM

PR #90, opened 2021-06-02
MSEARCH-129: Make consumer name different per environment.

bohdan-suprun:
Introduce `ENV` variable to differentiate environments and build consumer group id appending the env name.


hjiebsco (approved):

pfilippovEPAM (approved):
Looks good to me.

PR #91, opened 2021-06-03
MSEARCH-126 Optimize indexing process for field processors …

pfilippovEPAM:
### Purpose

The existing implementation contains a lot of search field processors with JsonConverter inside to extract map and convert map value to the instance field. It would be faster if the application can convert once map to the requested resource class and then the developer will be able to work resource as java class, not as an unstructured map.

### Approach

- All instance field processors are re-implemented to apply Instance object instead of Map
- Instance per event converted only once before further processing by the field processors
- Event body class can be specified with the eventBodyJavaClass property in the resource description
- Additional validation added for resources with eventBodyJavaClass property in the resource description
- Resource without eventBodyJavaClass requires processors with Map<String, Object> generic type
- Search field processor logic extracted to SearchFieldProcessor class
- Unit tests are re-implemented for all instance field processors
- An integration test for InstanceIdentifierTypeCache converted from integration to unit and its performance increased using specific Spring context (only caching and required beans instead of loading whole application)
- Spring boot tests are added for SearchFieldProcessor and ResourceDescriptionService


hjiebsco (commented):
Thanks for the refactoring effort. I will leave others to do a deep review.

src/main/java/org/folio/search/service/converter/SearchFieldProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.commons.collections.MapUtils;
+import org.folio.search.model.metadata.SearchFieldDescriptor;
+import org.folio.search.service.setter.FieldProcessor;
+import org.folio.search.utils.JsonConverter;
+import org.springframework.stereotype.Component;
+
+@Log4j2
+@Component
+@RequiredArgsConstructor
+public class SearchFieldProcessor {
```
bohdan-suprun, line 20:
```suggestion
public class SearchFieldProcessors {
```

Maybe add `s` to the class name?

src/main/java/org/folio/search/service/converter/SearchFieldProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @SuppressWarnings("rawtypes")
+  private final Map<String, FieldProcessor> fieldProcessors;
```
bohdan-suprun, line 23: Can it be changed to something like following? 
```suggestion
  private final Map<String, FieldProcessor<?>> fieldProcessors;
```

bohdan-suprun (approved):

src/main/java/org/folio/search/service/converter/SearchFieldProcessor.java
pfilippovEPAM: It's a really rare case to class to have a name in plural form, but I think, you have a point and I will rename it to the `SearchFieldsProcessor` to mark that is not a single field processor

PR #93, opened 2021-06-09
MSEARCH-95: Use ENV name as prefix for Elasticsearch index.

bohdan-suprun:
The `ENV` variable is used as a prefix for ES index, `ENV` value is restricted to contain alphanumeric and '_', '-' symbols only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ with less powerful configuration (see [High availability](https://www.elastic.co
 | INITIAL_LANGUAGES             | eng                       | Comma separated list of languages for multilang fields see [Multi-lang search support](#multi-language-search-support) |
 | SYSTEM_USER_PASSWORD          | -                         | Password for `mod-search` system user (not required for dev envs) |
 | OKAPI_URL                     | -                         | OKAPI URL used to login system user, required                     |
-| ENV                           | folio                     | Logical name of the deployment, must be set if Kafka/Elasticsearch are shared for environments |
+| ENV                           | folio                     | Logical name of the deployment, must be set if Kafka/Elasticsearch are shared for environments, `a-z (any case)`, `0-9`, `-`, `_` symbols only allowed|
 
 The module uses system user to communicate with other modules from Kafka consumers.
 For production deployments you MUST specify the password for this system user via env variable:
--- a/src/main/java/org/folio/search/configuration/properties/FolioEnvironment.java
+++ b/src/main/java/org/folio/search/configuration/properties/FolioEnvironment.java
@@ -2,22 +2,13 @@ package org.folio.search.configuration.properties;
 
 import static java.lang.System.getProperty;
 import static java.lang.System.getenv;
+import static lombok.AccessLevel.PRIVATE;
 import static org.apache.commons.lang3.StringUtils.firstNonBlank;
 
-import lombok.Getter;
-import org.springframework.beans.factory.annotation.Value;
-import org.springframework.stereotype.Component;
+import lombok.NoArgsConstructor;
 
-@Getter
-@Component
-public class FolioEnvironment {
-  private final String okapiUrl;
-  private final String env;
-
-  public FolioEnvironment(@Value("${okapi.url}") String okapiUrl) {
-    this.okapiUrl = okapiUrl;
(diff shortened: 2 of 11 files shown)


src/main/java/org/folio/search/configuration/properties/FolioEnvironment.java
```diff
@@ -10,10 +14,20 @@
...
-
+  public FolioEnvironment(@Value("${okapi.url}") String okapiUrl) {
     this.okapiUrl = okapiUrl;
-    this.env = env;
+    this.env = getFolioEnvName();
+  }
+
+  public static String getFolioEnvName() {
+    return validateFolioEnv(firstNonBlank(getenv("ENV"), getProperty("env"), "folio"));
+  }
```
hjiebsco, line 24: Why is this handled different from other docker environment variables?

src/main/java/org/folio/search/utils/SearchUtils.java
```diff
@@ -157,4 +158,8 @@ public static String getEffectiveCallNumber(String prefix, String callNumber, St
   public static <T> Stream<T> toSafeStream(List<T> nullableList) {
     return CollectionUtils.isNotEmpty(nullableList) ? nullableList.stream() : Stream.empty();
   }
+
+  private static String getFolioEnvAndMakeValidEsIdentifier() {
+    return getFolioEnvName().replace("-", "_").toLowerCase();
+  }
```
hjiebsco, line 164: Is "-" not allowed? The doc did not mention that explicitly https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-indexing.html

hjiebsco (commented):
Wondering why `ENV` (and `OKAPI_URL`) are handled differently from other environment variables specified in module descriptor.

README.md
```diff
@@ -123,7 +123,7 @@ with less powerful configuration (see [High availability](https://www.elastic.co
 | INITIAL_LANGUAGES             | eng                       | Comma separated list of languages for multilang fields see [Multi-lang search support](#multi-language-search-support) |
 | SYSTEM_USER_PASSWORD          | -                         | Password for `mod-search` system user (not required for dev envs) |
 | OKAPI_URL                     | -                         | OKAPI URL used to login system user, required                     |
-| ENV                           | folio                     | Logical name of the deployment, must be set if Kafka/Elasticsearch are shared for environments |
+| ENV                           | folio                     | Logical name of the deployment, must be set if Kafka/Elasticsearch are shared for environments, `a-z (any case)`, `0-9`, `-`, `_` symbols only allowed|
```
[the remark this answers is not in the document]
bohdan-suprun: I'd not disallow it, it does not make any harm I think.

src/main/java/org/folio/search/configuration/properties/FolioEnvironment.java
bohdan-suprun: Because logic that builds index name is static (see `org.folio.search.utils.SearchUtils#getElasticsearchIndexName(java.lang.String, java.lang.String)`), it is the easiest way to implement it without refactoring to be non-static.

src/main/java/org/folio/search/utils/SearchUtils.java
bohdan-suprun: You're right, removed this method.

src/main/java/org/folio/search/configuration/properties/FolioEnvironment.java
```diff
@@ -10,10 +14,20 @@
   private final String okapiUrl;
   private final String env;
 
-  public FolioEnvironment(@Value("${okapi.url}") String okapiUrl,
-    @Value("${env:folio}") String env) {
-
+  public FolioEnvironment(@Value("${okapi.url}") String okapiUrl) {
```
pfilippovEPAM, line 17: can we mover okapi.url to the configuration properties like OkapiConfigurationProperties

pfilippovEPAM (approved):
Looks good to me, except one comment, but it can be ignored

MikhailFokanov (approved):
Looks good to me. But you could address Pavel's comments.

PR #96, opened 2021-06-22
MSEARCH-134 Remove multi-language support from the contributors fields

pfilippovEPAM:
### Purpose
While searching contributors =="carols" the search returns records with contributors Carole and Carol

### Acceptance criteria:
* in searches for `contributors == "<name>s"` is not treated as a plural form but as part of the name

### Approach

The main change is adding support for fields with index = `standard` as full-text fields. They will be recognized by the CQL query converter and the same rules as for the multilanguage fields will be applied.

Integration and unit tests are added to prove of added functionality

### DIFF
--- a/src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
+++ b/src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
@@ -14,7 +14,7 @@ import static org.elasticsearch.index.query.QueryBuilders.wildcardQuery;
 import static org.folio.search.utils.SearchQueryUtils.isBoolQuery;
 import static org.folio.search.utils.SearchQueryUtils.isDisjunctionFilterQuery;
 import static org.folio.search.utils.SearchQueryUtils.isFilterQuery;
-import static org.folio.search.utils.SearchUtils.updatePathForMultilangField;
+import static org.folio.search.utils.SearchUtils.updatePathForFulltextField;
 import static org.folio.search.utils.SearchUtils.updatePathForTermQueries;
 
 import java.io.IOException;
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/search/cql/CqlSearchQueryConverter.java
```diff
@@ -251,8 +251,8 @@ private String getSearchTerm(String term, String fieldName, String resource) {
 
   private List<String> getFieldsForMultilangField(String fieldName, String resource) {
```
MikhailFokanov-ebsco, line 252: Why .filter(PlainFieldDescription::isFulltex  is used in getFieldsForMultilangField?

src/main/java/org/folio/search/model/metadata/PlainFieldDescription.java
```diff
@@ -66,19 +70,47 @@
...
+    return STANDARD_FIELD_TYPE.equals(index);
+  }
+
+  /**
+   * Checks if resource description field is full-text.
+   *
+   * @return true if field is standard full text, false - otherwise
+   */
+  @JsonProperty
+  public boolean isFulltext() {
```
MikhailFokanov-ebsco, line 110: unclear naming, because field isFulltext is likely to be read as getter of "fulltext" private field. Or can be read similar to others in this class.

src/main/java/org/folio/search/model/metadata/PlainFieldDescription.java
```diff
@@ -66,19 +70,47 @@
...
     return MULTILANG_FIELD_TYPE.equals(index);
   }
 
+  /**
+   * Checks if resource description field is standard full-text.
+   *
+   * @return true if field is standard full text, false - otherwise
+   */
+  @JsonIgnore
+  public boolean isStandardFulltext() {
```
MikhailFokanov-ebsco, line 100: please rename to isStandard?

MikhailFokanov (approved):

PR #97, opened 2021-06-30
MSEARCH-132 Apply new topic naming convention

pfilippovEPAM:
### Purpose

Many modules that use Kafka use ENV variable and tenant id in their naming convention for topics.

For example topics in Bugfest have names that include ENV (ibf3) in addition to tenant id (fs09000000) - such as the following topic example

`ibf3.Default.fs09000000.DI_COMPLETED`

It would be helpful if ENV and tenant id settings were used in the topic naming convention. This would allow a Kafka instance to be shared by multiple environments that have the same tenant id.

https://github.com/folio-org/mod-search/blob/master/src/main/resources/kafka/kafka-topics.json

Required topic pattern - `{env}.{tenant}.{topicname}`

### Approach

A new approach listens to the topics patterns instead of a list of provided topics. This will allow to mod-search receive messages from topics starting from the `ENV` variable and following any tenant name. A batch request to Elasticsearch will be formed for all tenants, but data will be separate at different indices.

### DIFF
--- a/src/main/java/org/folio/search/configuration/properties/FolioKafkaProperties.java
+++ b/src/main/java/org/folio/search/configuration/properties/FolioKafkaProperties.java
@@ -3,11 +3,13 @@ package org.folio.search.configuration.properties;
 import java.util.Map;
 import lombok.Data;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
 
 /**
  * Application properties for kafka message consumer.
  */
 @Data
+@Component
 @ConfigurationProperties("application.kafka")
 public class FolioKafkaProperties {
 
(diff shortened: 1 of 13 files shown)


src/main/java/org/folio/search/configuration/properties/FolioKafkaProperties.java
```diff
@@ -26,6 +28,16 @@
    */
   private long retryDeliveryAttempts;
 
+  /**
+   * List of default topics to maintain backward compatibility.
+   */
+  private String topicsRegexp;
```
MikhailFokanov, line 34: The name `topicsRegexp` is missleading, as you also have `topicPattern`. If it is legacy, let's name it `legacyTopicPattern` or `defaultTopicPattern` and mark as @Deprecated

pfilippovEPAM: Yes, thanks for noticing this, this change is not required at all

MikhailFokanov (approved):

PR #98, opened 2021-06-30
MSEARCH-136 Adjust sorting by title to use indexTitle if possible

pfilippovEPAM:
### Purpose

The current implementation sorts by `instance.title` property but the preferred behavior is to sort by `instance.indexTitle`

### Requirements/Scope
The sorting by title should behave as follows:

- if `instance.indexTitle` is populated, it is used for sorting
- if `instance.indexTitle` is not populated, sort by `instance.title` field

### Approach

Implement new `SortTitleProcessor` which will chose one of titles - `indexTitle` or `title`. If `indexTitle` is blank, `title` will be analyzed and if it is not empty - it will be used for sorting. If `indexTitle` is not blank - it will be used for sorting.
Integration and unit tests are added to prove developed functionality.

### DIFF
--- a/src/main/java/org/folio/search/service/setter/instance/SortTitleProcessor.java
+++ b/src/main/java/org/folio/search/service/setter/instance/SortTitleProcessor.java
@@ -1,0 +1,20 @@
+package org.folio.search.service.setter.instance;
+
+import static org.apache.commons.lang3.StringUtils.defaultIfBlank;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+
+import lombok.RequiredArgsConstructor;
+import org.folio.search.domain.dto.Instance;
+import org.folio.search.service.setter.FieldProcessor;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public final class SortTitleProcessor implements FieldProcessor<Instance, String> {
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/search/service/setter/instance/SortTitleProcessor.java
```diff
@@ -0,0 +1,20 @@
...
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+
+import lombok.RequiredArgsConstructor;
+import org.folio.search.domain.dto.Instance;
+import org.folio.search.service.setter.FieldProcessor;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public final class SortTitleProcessor implements FieldProcessor<Instance, String> {
```
MikhailFokanov, line 13: Maybe name it as TreatEmptyAsNullProcessor to allow usage of it in another cases?

pfilippovEPAM: All mod-search processors have the same naming convention, I guess, we can create an abstract class if this behaviour will repeat again, but not now

MikhailFokanov (approved):

PR #99, opened 2021-07-07
MSEARCH-117 Support full text search for notes of linked holding-records

pfilippovEPAM:
### Purpose
The purpose of this pull request is to allow the user a full-text search within the instance/item/holdings record's notes field. Search should support using public only and all `notes.note` values.

### Acceptance criteria

- Existing search by public notes of instance only is preserved
- Additional search option notes (all) is added
- Query returns accurate hit count
- User can search by any word that is a part of the field's value
- Results are ranked based on term frequency, inverse document frequency, and field length
- Stemming according to with configured on the tenant language tokenizers

### Approach
Since the approach for implementing full-text search support for the following fields `notes.note`, `items.notes.note`, `holdings.notes.note` is similar, all 3 stories are merged into a single pull request.
Integration and unit test are added to prove implemented functionality

### Related stories, that are implemented within this pull request:
- [MSEARCH-116 Instance - support full-text search notes field](https://issues.folio.org/browse/MSEARCH-116)
- [MSEARCH-118 Item- support full-text search notes field](https://issues.folio.org/browse/MSEARCH-118)

### DIFF
--- a/src/main/java/org/folio/search/service/setter/holding/HoldingPublicNotesProcessor.java
+++ b/src/main/java/org/folio/search/service/setter/holding/HoldingPublicNotesProcessor.java
@@ -1,0 +1,24 @@
+package org.folio.search.service.setter.holding;
+
+import static org.folio.search.utils.SearchUtils.toSafeStream;
+
+import java.util.Collection;
+import java.util.stream.Stream;
+import org.apache.commons.collections.CollectionUtils;
+import org.folio.search.domain.dto.Holding;
+import org.folio.search.domain.dto.Instance;
+import org.folio.search.domain.dto.Note;
+import org.folio.search.service.setter.instance.AbstractPublicNotesProcessor;
+import org.springframework.stereotype.Component;
+
+@Component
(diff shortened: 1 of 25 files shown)


src/main/java/org/folio/search/service/setter/holding/HoldingPublicNotesProcessor.java
```diff
@@ -0,0 +1,24 @@
...
+import org.folio.search.domain.dto.Note;
+import org.folio.search.service.setter.instance.AbstractPublicNotesProcessor;
+import org.springframework.stereotype.Component;
+
+@Component
+public class HoldingPublicNotesProcessor extends AbstractPublicNotesProcessor {
+
+  @Override
+  protected Stream<Note> getNotes(Instance instance) {
+    return toSafeStream(instance.getHoldings())
```
predictron-cloud, line 19: Maybe `toSafeStream` should be in `CollectionUtils.java`?

predictron-cloud (approved):

MikhailFokanov (approved):
Maybe toSafeStream should be in CollectionUtils.java?

PR #103, opened 2021-07-19
MSEARCH-155 Handle gracefully message processing failures

pfilippovEPAM:
### Purpose
If the notification message cannot be processed and the batch saving is failed. The batch should be divided into single messages and process one by one. For each of these single message processing, there should be a retried 3 times in case of error. After 3 unsuccessful retries, the message should be logged.
This approach allows mod-search to avoid "poison pill" behavior when one "bad" message can stop the whole system.

### Approach
- Implement FolioMessageBatchProcessor which will allow to retry batch n-times and then split the whole batch into a single message and process them one by one.
- Add integration and unit tests to prove functionality
- Retry bean can be specified for FolioMessageBatchProcessor so developers can use different strategies for batch saving for further developed Kafka message listeners.


src/main/java/org/folio/search/configuration/KafkaConfiguration.java
```diff
@@ -45,17 +44,22 @@
...
+   *
+   * @return created {@link RetryTemplate} object
    */
-  private BatchErrorHandler kafkaFactoryErrorHandler() {
-    return new RecoveringBatchErrorHandler(new FixedBackOff(
-      folioKafkaProperties.getRetryIntervalMs(), folioKafkaProperties.getRetryDeliveryAttempts()));
+  @Bean(name = KAFKA_RETRY_TEMPLATE_NAME)
+  public RetryTemplate kafkaMessageListenerRetryTemplate() {
+    return RetryTemplate.builder()
+      .maxAttempts((int) folioKafkaProperties.getRetryDeliveryAttempts())
```
MikhailFokanov, line 60: Absence of the value in case of casting to (int) would cause NullPointer? Can it be casted to (Integer)?

pfilippovEPAM: By default - this value is equal to 10 in the FolioKafkaProperties class, just in case, if no configuration has been provided.

MikhailFokanov (approved):

PR #105, opened 2021-07-23
MSEARCH-148 Support search by uniform title

pfilippovEPAM:
### Purpose/Overview

The current implementation of the title search includes searches by `title`, `alternativeTitle`, `indexTitle` and `series`.  There are, however, use cases that the librarian would like to search by the specified uniform title.  This is especially important for music libraries.

### Requirements/Scope:

- Add search option that will support uniform title search
- The search will occur in the `alternativeTitle` field  when the title type is set to Uniform title
- Not all records will have the uniform titles populated.
- The key to the Uniform title should be retrieved by the call to /alternative-title-types.  Since this is a user-specified value - it is possible that it will not be present - this needs to be handled gracefully as well
- The search by uniform title should follow the same rules as search by title

### Approach

- Implement `UniformTitleProcessor` which will be responsible for extracting uniform titles from alternative titles array
- Replace persistent cache with TTL because the uniform title is librarian specified value and can be created after mod-search initialization
- Implement unit and integration tests to prove implemented functionality
- Rename `InstanceIdentifierTypeCache ` to `InstanceReferenceDataService` because now it's responsible for all reference data, not only for identifier ids
- Fix incorrect mappings for `source` field
- Exclude all empty values from the index - it was an unexpected side effect because `SearchFieldsProcessor` checks output value only for `null`, now it also checks for empty map and collection


MikhailFokanov (approved):

MikhailFokanov:
Looks good. Thanks!

src/main/java/org/folio/search/utils/SearchUtils.java
```diff
@@ -200,4 +201,19 @@ public static String getEffectiveCallNumber(String prefix, String callNumber, St
     return fulltextValue;
   }
 
+  /**
+   * Checks if passed value null or empty collection or map.
+   *
+   * @param object - value to check
+   * @return true if null or empty, false - otherwise.
+   */
+  public static boolean isNotNullOrEmpty(Object object) {
```
OleksiiKuzminov, line 210: Is it possible to use Apache commons for such kind of checks?

OleksiiKuzminov (approved):
Looks good! Just one question

src/main/java/org/folio/search/utils/SearchUtils.java
pfilippovEPAM: resolved, I've replaced it with `ObjectUtils.isNotEmpty(Object value)`
thanks for noticing this!

PR #108, opened 2021-07-28
MSEARCH-160 Failed to create index: Limit of total fields [250] has been exceeded

OleksiiKuzminov:
### Purpose/Overview:

The current configuration has a value for the number of fields limitation as 250. With 5 assigned languages it raised an error for the previous version of mapping. 

### Approach
- Change mapping.total_fields.limit property to 1000
- Add 5th language to the test configuration to cover the maximum number of fields


pfilippovEPAM (approved):
Looks good to me. Thank you

PR #110, opened 2021-08-13
MSEARCH-152/153 Add Identifiers (all) search option for holding and items

OleksiiKuzminov:
### Purpose/Overview:

The current implementation supports holdings search by HRID but librarians also need to search by holdings and items by their former Id.  While searching it is not always clear if the provided value is a current or a former HRID.  For those cases, there need to be an option to cover both in one search

### Requirements/Scope:

- Add Identifiers (all) option for holdings and items
- The search should include matches from hrid and fromerIds (array of strings)

### Approach
-Added new search fields: itemIdentifiers and holdingIdentifiers. These fields should support searching by hrId and former ids

### DIFF
--- a/src/test/java/org/folio/search/controller/SearchInstanceIT.java
+++ b/src/test/java/org/folio/search/controller/SearchInstanceIT.java
@@ -29,7 +29,7 @@ class SearchInstanceIT extends BaseIntegrationTest {
   @MethodSource("positiveSearchTestDataProvider")
   @DisplayName("can search by instances (index with only 1 instance)")
   void canSearchByInstances(String testName, String query, Object[] queryArguments,
-    ThrowingConsumer<ResultActions> extendedSearchResultMatcher) throws Throwable {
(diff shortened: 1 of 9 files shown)


src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -29,7 +29,7 @@
   @MethodSource("positiveSearchTestDataProvider")
   @DisplayName("can search by instances (index with only 1 instance)")
   void canSearchByInstances(String testName, String query, Object[] queryArguments,
-    ThrowingConsumer<ResultActions> extendedSearchResultMatcher) throws Throwable {
+                            ThrowingConsumer<ResultActions> extendedSearchResultMatcher) throws Throwable {
```
pfilippovEPAM, line 32: I think that changes is not needed here

pfilippovEPAM (commented):
Looks good to me except one comment

pfilippovEPAM (approved):

PR #116, opened 2021-08-17
MSEARCH-171 Update Dockerfile to use folioci java Docker image

pfilippovEPAM:
### Purpose
Using the existing Dockerfile configuration is not efficient and does not allow to unify environment variables like `JAVA_OPTIONS`. Also, other folio Spring-based modules are using `folioci/alpine-jre-openjdk11:latest` as the root Docker image.

### Approach
- Update `Dockerfile` to use `folioci/alpine-jre-openjdk11:latest` as the root Docker image
- Remove redundant `run.sh` file
- Update Docker compose file to run the application locally using Docker image
- Add to docker-compose PostgreSQL as service

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,12 +1,22 @@
 version: "3.8"
 
 services:
+  mod-search:
+    container_name: mod-search
+    image: dev.folio/mod-search:latest
+    ports:
+      - "8081:8081"
+    environment:
+      DB_HOST: db
+      JAVA_OPTIONS: -Xmx120m -Xms120m
+      okapi.url: http://okapi:9130
+
   elasticsearch:
     container_name: elasticsearch
-    image: org.folio/elasticsearch:latest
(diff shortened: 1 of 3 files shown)


docker/docker-compose.yml
```diff
@@ -51,5 +61,16 @@ services:
...
+    image: postgres:13.4-alpine
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_DB: okapi_modules
+
 volumes:
-  es-data: { }
```
MikhailFokanov-ebsco, line 55: Is it ok not to have newline here?

pfilippovEPAM: this line is removed, so yes
after adding persistent volume for PostgreSQL  `docker-compose` file looks good

OleksiiKuzminov (approved):

MikhailFokanov-ebsco (approved):

PR #118, opened 2021-08-18
MSEARCH-130 align dependency versions affected by Inventory's Optimistic Locking

MikhailFokanov:
MSEARCH-130 align dependency versions affected by Inventory's Optimistic Locking


descriptors/ModuleDescriptor-template.json
```diff
@@ -131,7 +131,7 @@
   "requires": [
     {
       "id": "instance-storage",
-      "version": "7.5"
+      "version": "8.0"
```
julianladisch, line 134: If there is not code change in mod-search it supports both 7.5 and 8.0 and therefore both should be listed.

julianladisch: Listing both is also needed to merge this pull request before merging the mod-inventory-storage pull request.

julianladisch: - [x] Done, thanks!

julianladisch:
Can you check whether the _version property should be added to these files?
src/main/resources/swagger.api/schemas/instance.json
src/main/resources/swagger.api/schemas/holding.json
src/main/resources/swagger.api/schemas/item.json
src/test/resources/samples/semantic-web-primer/instance.json
src/test/resources/samples/semantic-web-primer/items.json
src/test/resources/samples/semantic-web-primer/holdings.json

MikhailFokanov:
@julianladisch, Search have not full schema of inventory some fields are not needed and hence ignored. The version column is also not needed, as there is no writes in mod-search, so as far as I can see I don't need to change schemas and tests.

julianladisch (approved):
Looks good, thanks!

PR #121, opened 2021-08-24
MSEARCH-173 Decrease the retry timeout

pfilippov-epam:
### Purpose/Overview:

Current retry settings for failed messages are

```
application:
  kafka:
    retry-interval-ms: 20000
    retry-delivery-attempts: 10
```
This leads that initial bulk processed with a time equal to:

```10 * 20 sec = 200 sec ```
and then, if it's failing every message is processed with the same amount of time. So the gracefully handled bulk processed with time:

```200s + 50 (amount of messages in bulk) * 10 * 20 sec = 10200 sec = 170 min```

This value must be decreased to 10 minutes to not stop the processing of messages for a long time. The current implementation will log every unprocessed message.

### Requirements/Scope:

Kafka settings: retry-interval-ms and retry-delivery-attempts must be decreased
- `retry-interval-ms` must be set to 2000
- `retry-delivery-attempts` must be set to 6
- Make these options configurable by setting environment variables: KAFKA_RETRY_INTERVAL_MS and KAFKA_RETRY_DELIVERY_ATTEMPTS

### Approach
- Ignore events when related Elasticsearch index does not exist (they won't be processed anyway), but they will be still logged to maintain observability
- Update configuration properties

### DIFF
--- a/src/main/java/org/folio/search/service/IndexService.java
+++ b/src/main/java/org/folio/search/service/IndexService.java
@@ -3,6 +3,7 @@ package org.folio.search.service;
 import static java.lang.Boolean.TRUE;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
 import static org.folio.search.model.types.IndexActionType.DELETE;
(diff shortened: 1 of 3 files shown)


MikhailFokanov (approved):
Looks good to me, thanks.

PR #122, opened 2021-09-07
MSEARCH-174 Investigate indexing issues in reference environment

OleksiiKuzminov:
### Purpose/Overview:

Currently, if mod-search is initialized for the tenant where DB already has instances, the user can’t find these instances. 

### Requirements/Scope:

- new tenant attribute processing should be added
- on _**runReindex**_ modules should start reindex

### Approach
New tenant attributes can be set up on environments with stub data which is created before mod-search starts. _**runReindex**_ attribute should trigger the full reindex on module start up

### DIFF
--- a/src/main/java/org/folio/search/controller/FolioTenantController.java
+++ b/src/main/java/org/folio/search/controller/FolioTenantController.java
@@ -1,8 +1,14 @@
 package org.folio.search.controller;
 
+import static org.folio.search.utils.SearchUtils.INSTANCE_RESOURCE;
+
+import java.util.Collection;
+import java.util.stream.Stream;
 import lombok.extern.log4j.Log4j2;
+import org.folio.search.service.IndexService;
 import org.folio.search.service.KafkaAdminService;
 import org.folio.search.service.SearchTenantService;
+import org.folio.spring.FolioExecutionContext;
 import org.folio.spring.controller.TenantController;
 import org.folio.spring.service.TenantService;
 import org.folio.tenant.domain.dto.TenantAttributes;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/search/controller/FolioTenantController.java
```diff
@@ -36,6 +35,7 @@ public FolioTenantController(TenantService baseTenantService,
 
     if (tenantInit.getStatusCode() == HttpStatus.OK) {
       tenantService.initializeTenant();
+      tenantService.reIndexInstances(tenantAttributes);
```
pfilippov-epam, line 38: why it's cannot be part of the method - `initializeTenant()`?

src/main/java/org/folio/search/service/SearchTenantService.java
```diff
@@ -46,4 +50,11 @@ public void removeElasticsearchIndexes() {
       indexService.dropIndex(resource.getName(), context.getTenantId());
     }
   }
+
+  public void reIndexInstances(TenantAttributes tenantAttributes) {
+    Stream.ofNullable(tenantAttributes.getParameters()).flatMap(Collection::stream)
+      .filter(parameter -> parameter.getKey().equals(REINDEX_PARAM_NAME)
+        && Boolean.parseBoolean(parameter.getValue()))
```
pfilippov-epam, line 57: I would use static import for it

pfilippov-epam (commented):
Looks good, except a few minor comments

pfilippov-epam (approved):

MikhailFokanov (approved):
Looks good.

PR #131, opened 2021-09-29
MSEARCH-184 Implement streaming of holding ids

pfilippov-epam:
### Purpose
Similar to streaming instances UUIDs, there is a need to implement streaming of holdings UUIDs so that the user can identify a list of holdings records that match the search criteria. This endpoint will be used in the Inventory app to allow users to save holding UUIDs that can be used later as a base for holdings record export

### Requirements
Add a new endpoint in mod-search to stream holdings UUIDs

### Acceptance criteria:

- Only holdings UUIDs are returned
- Support saving UUIDs for at least 1M records

### Approach

- Implement new endpoint `search/holdings/ids` with required parameter `query`.
- Extract duplicated code to the `ResourceIdsStreamHelper` class
- Update resource ids query to accept source variable from request and extract related values from response\
- Add integration test case for holdings ids streaming feature


OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good.

PR #132, opened 2021-09-30
Update for mod-search.yaml

OleksiiKuzminov:
### Purpose
Update a mod-search.yaml and lead to common standards

### Approach
- unwrap query parameters and remove refs


src/main/java/org/folio/search/mapper/SearchRequestMapper.java
```diff
@@ -13,24 +12,29 @@
...
+   * @param offset  skip over a number of elements by specifying an offset value for the query {@link Integer}
+   * @param expandAll  whether to return only basic properties or entire instance {@link Boolean}
    * @param resource resource name as {@link String}
    * @param tenantId tenant id as {@link String}
    * @return created {@link CqlSearchServiceRequest} object
    */
   @Mapping(source = "resource", target = "resource")
   @Mapping(source = "tenantId", target = "tenantId")
-  CqlSearchServiceRequest convert(CqlSearchRequest request, String resource, String tenantId);
+  CqlSearchServiceRequest convert(String query, Integer limit, Integer offset, Boolean expandAll, String resource,
```
pfilippov-epam, line 25: I guess there is no need to keep in that case SearchRequestMapper.

pfilippov-epam (approved):
Looks good except one comment, but it can be ignored for now and we can remove SearchRequestMapper in separate story

PR #137, opened 2021-10-05
MSEARCH-175 Item - include circulation notes search in item notes search

OleksiiKuzminov:
### Requirements/Scope:

- If the circulation note is mark as staff only, it is included in notes (all) search but is not included in public notes search results
- if the circulation note is not marked as staff only, it is included in public notes search results
- the functionality introduced in MSEARCH-158 stays as it is and the user can also search for circulation notes separately

### DIFF
--- a/src/main/java/org/folio/search/service/setter/instance/AbstractPublicNotesProcessor.java
+++ b/src/main/java/org/folio/search/service/setter/instance/AbstractPublicNotesProcessor.java
@@ -1,10 +1,13 @@
 package org.folio.search.service.setter.instance;
 
 import static java.util.stream.Collectors.toCollection;
+import static java.util.stream.Collectors.toList;
 
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Stream;
 import org.folio.search.domain.dto.CirculationNote;
 import org.folio.search.domain.dto.Instance;
@@ -20,14 +23,13 @@ public abstract class AbstractPublicNotesProcessor implements FieldProcessor<Ins
     result.addAll(getNotesAsList(getCirculationNotes(instance), note -> getNote(note.getStaffOnly(), note.getNote())));
     return result;
   }
-  
+
   private static <T> List<String> getNotesAsList(Stream<T> notesStream, Function<T, String> func) {
     return notesStream.filter(Objects::nonNull).map(func).filter(Objects::nonNull).collect(toList());
   }
 
   private static String getNote(Boolean staffOnly, String value) {
     return staffOnly == null || !staffOnly ? value : null;
-  }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -159,10 +159,16 @@ void streamInstanceIds() throws Exception {
...
+      arguments("search by public circulation notes items.notes.note", "itemPublicNotes all {value}",
+        array("first-record"), zeroResultConsumer()),
+      arguments("search by public circulation notes items.notes.note", "itemPublicNotes all {value}",
+        array("SecondRecord"), null),
       arguments("search by private items.notes.note using itemPublicNotes", "itemPublicNotes == {value}",
         array("librarian private note for item"), zeroResultConsumer()),
       arguments("search by private items.notes.note", "items.notes.note == {value}",
         array("Librarian private note for item"), null),
+      arguments("search by circulation notes in items.notes.note", "items.notes.note == {value}",
+        array("testNote"), null),
```
pfilippov-epam, line 171: can you add a test case that search for `first-record` will return correct result using query
```
items.note.note == "first-record"
```

src/main/java/org/folio/search/service/setter/item/ItemPublicNotesProcessor.java
```diff
@@ -21,4 +22,12 @@
       .filter(CollectionUtils::isNotEmpty)
       .flatMap(Collection::stream);
   }
+
+  @Override
+  protected Stream<CirculationNote> getCirculationNotes(Instance instance) {
```
pfilippov-epam, line 27: Can you add one more case into a related unit test for the item with notes and circulationNotes?

src/main/java/org/folio/search/service/setter/instance/AbstractPublicNotesProcessor.java
```diff
@@ -14,12 +15,28 @@
...
+        .filter(Objects::nonNull)
+        .filter(note -> note.getStaffOnly() == null || !note.getStaffOnly())
+        .map(CirculationNote::getNote)
+        .filter(Objects::nonNull),
+      getNotes(instance).filter(Objects::nonNull)
+        .filter(note -> note.getStaffOnly() == null || !note.getStaffOnly())
+        .map(Note::getNote)
+        .filter(Objects::nonNull)
+    ).collect(toCollection(LinkedHashSet::new));
+
```
pfilippov-epam, line 29:
```suggestion
    var result = new LinkedHashSet<String>();
    result.addAll(getNotesAsList(getNotes(instance), note -> getNote(note.getStaffOnly(), note.getNote())));
    result.addAll(getNotesAsList(getCirculationNotes(instance), note -> getNote(note.getStaffOnly(), note.getNote())));
    return result;
  }
  
  private static <T> List<String> getNotesAsList(Stream<T> notesStream, Function<T, String> func) {
    return notesStream.filter(Objects::nonNull).map(func).filter(Objects::nonNull).collect(toList());
  }

  private static String getNote(Boolean staffOnly, String value) {
    return staffOnly == null || !staffOnly ? null : value;
  }
```

PR #140, opened 2021-10-18
MSEARCH-199 Fix facet for instanceFormatIds

pfilippov-epam:
### Purpose
Filtering of Format has stopped working after the swap to use ES as a search engine - work done UIIN-1567. Format should be working similarly to e.g. Resource type, except that the values are concatenated Format type, format value.

### Approach
- Fix invalid mappings name
- Fix integration tests cases to use the correct value - `instanceFormatIds`
- Update README.md documentation


MikhailFokanov (approved):
Looks good

PR #142, opened 2021-10-27
MSEARCH-195 Implement search by authority records

pfilippov-epam:
### Purpose
To be able to create a mapping and search by authority record (AR), it needs to have an API, Model, and DTO in mod-search

### Requirements/Scope:
- Add AR model and Dto (simple mapping id, name)
- Add new SearchResource
- Extend `mod-search.yaml` with new API `/search/authorities`
- Add Kafka Handler(separate consumer 1-thread) to receive messages from inventory-storage
- Add unit/integration tests

### Approach
- Add AR description - model JSON, index settings.
- Improve approach to receive messages from Kafka directly and index them directly to Elasticsearch
- Unify search response from SearchService (controllers can convert it to the object generated by `OpenAPI`)
- Update `FolioTenantController` to be able to create and delete indices for instance and AR
- Integration tests
  - Add samples for searching
  - Add integration test for AR index with a single record
  - Refactor BaseIntegrationTest to be able to setUp tenant for AR-related tests

### Additional changes
- Update indexing approach for resources loaded directly from Kafka
- Fix issue with wildcard searching for fields uses standard mapping with plain value
- Fix mapping creation for standard fields with plain value (since by default `personalName` is indexed with the same approach as the `contributors.name` field from `instance`)
- Update Kafka topic creation approach - restarting all event listeners on enabling tenants, Kafka creation approach, defining replication factor directly from application env variables
- Refactor integration tests - update setUp tenant operations, search endpoints, use static mockMvc reference for all operations


OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good.

PR #143, opened 2021-11-01
MSEARCH-192 Add formerIds field for search by all item fields

pfilippov-epam:
### Purpose
Search by query `cql.all` or `cql.allItems` does not provide results for item field - `formerIds`. It's happening because by initial implementation all fields ending with `Id` or `Ids` are ignored. 

### Appraoch
- Add `formerIds` field as indexed during collecting field for queries `cql.all` or `cql.allItems`
- Update integration and unit tests


OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good

PR #145, opened 2021-11-03
FOLSPRINGB-4 Fix Maven replacement tokens in descriptor templates

pfilippov-epam:
### Purpose 
The replacement tokens for Spring-based modules are different from the normal form.

Instead, they require tokens of the form: "@artifactId@-@version@"

The maven build then transforms the replacement tokens (for "artifactId" and "version") in those template descriptors, and the resultant final Descriptors are generated to "./target/ModuleDescriptor.json" etc.

### Approach
- Use `@artifactId@-@version@` instead of `${artifactId}-${version}`
- Fix versions for API and json formatting of module descriptor template file

### DIFF
--- a/descriptors/DeploymentDescriptor-template.json
+++ b/descriptors/DeploymentDescriptor-template.json
@@ -1,5 +1,5 @@
 {
-  "srvcId": "${artifactId}-${version}",
+  "srvcId": "@artifactId@-@version@",
   "nodeId": "localhost",
   "descriptor": {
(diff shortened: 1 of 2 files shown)


OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good.

dcrossleyau (approved):
Okay. (I only investigated the changes for the tokens.)

PR #146, opened 2021-11-08
MSEARCH-196 Refactor reindex API to support authority

pfilippov-epam:
### Purpose
Change existing /reindex API to support not only Instances. 
Add resources field to support other resource types.

### Approach
- Implement and refactor the existing approach to accept other resources to reindex using as a base endpoint for reindexing following pattern - `{type}-storage/reindex`.
- Add integration tests to check that reindex service sends requests to the correct endpoint.


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good.

PR #147, opened 2021-11-08
Increase memory limits

pfilippov-epam:
### Purpose
Increase memory limits for the mod-search to prevent crushing by out of memory

### Approach 
- Increase memory limits in ModuleDescriptor-template.json


MikhailFokanov (approved):
I could try it...

PR #150, opened 2021-11-15
MSEARCH-210 Implement searching by divided authorities

pfilippov-epam:
### Purpose
To be able to determine the type of heading for authority record all incoming authority events must be divided into multiple records using criterion - distinctiveType from resource description.
The results list will need to include the type of heading that was used to determine the search match so that the user can easily identify the type of record that provides authority information

### Approach
- Implement an approach to divide single authority event into multiple using resource field description value - `distinctiveType`
- Add unit and integration tests to prove the approach
- Implement population of field - `headingType` based on the authority record field values
- Check that filtering by this field is working using integration tests
- Rename `ResourceEventBody` class into `ResourceEvent` to unify it with other class names


src/main/java/org/folio/search/model/types/FieldType.java
```diff
@@ -21,7 +21,13 @@
...
    * Additional field that to be generated by a processor.
    */
-  SEARCH("search");
+  SEARCH("search"),
+
+  /**
+   * Field designed for splitting entity to the multiple entities using that the value of the field as a criterion to
+   * split.
+   */
+  DISTINCTIVE("distinctive");
```
MikhailFokanov, line 30: Maybe splitted ?

MikhailFokanov (approved):
Looks good.

MikhailFokanov (approved):
looks good.

OleksiiKuzminov (approved):

PR #152, opened 2021-11-17
MSEARCH-207 Saving large set of UUIDs fails

OleksiiKuzminov:


MikhailFokanov-ebsco (approved):
I think retries could be the right solution for bug and also they wouldn't be harmful in any case.

pfilippov-epam (approved):
Looks good

PR #153, opened 2021-11-24
MSEARCH-212 Implement Authorized/Reference field

pfilippov-epam:
### Purpose
The results list will need to include the information that will specify where the match was found so that the user can easily identify the type of record that provides authority information. The heading type must be correctly identified based on the source of the match.

### Approach
- Implement new subfield for authority field description - `authRefType`
- Implement search field `authRefType` with corresponding field processor which will populate this field with related values for further filtering
- Add a single integration test to check that new field can be used as a filter


OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good.

PR #159, opened 2021-11-25
MSEARCH-211 Implement Heading/Reference field

pfilippov-epam:
### Purpose
The results list will need to include the information that will specify where the match was found so that the user can easily identify the Heading/Reference of record that provides authority information.

Approach
* Implement new subfield for authority field description - `headingRef`
* This field is currently stored as `source` and it can not be searchable using queries like - `headingRef all John`
* Add a single integration test to check that new field can be used as a filter


OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good.

PR #161, opened 2021-11-25
MSEARCH-234 Authority record search - Date created/updated filters

Shans-Kaluhin:
### Purpose/Overview:
 The user should have an option to filter authority records by time the record was created or updated. 

### Requirements/Scope:

Following fields provide filter search:
- The user can narrow search by specifying start and end date when the record was created  based on metadata.createdDate field (type: string, format: date-format) 
- The user can narrow search by specifying start and end date when the record was updated based on metadata.updatedDate field (type: string, format: date-format)
- The provided date should support year, month and day  (YYYY-MM-DD)
- Search should provide start date and end date (from, to) but the end date should not be required

### Acceptance criteria:
* Search returns accurate hit count
* Filter can be combined with other filters and represent logical conjunction
* User can search for records that were created starting with yyyy-mm-dd  -*todo: verify the expected date format*
* User can search for records that were created between yyyy-mm-dd and yyyy-mm-dd
* User can search for records that were updated starting with yyyy-mm-dd
* User can search for records that were updated between yyyy-mm-dd and yyyy-mm-dd


src/main/resources/model/authority.json
```diff
@@ -164,22 +164,24 @@
       "type": "object",
       "properties": {
         "createdDate": {
-          "index": "source"
+          "searchTypes": "filter",
+          "index": "date"
         },
         "createdByUserId": {
-          "index": "source"
+          "index": "keyword"
```
pfilippov-epam, line 171: changing the mapping for the fields:
- createdByUserId
- createdByUsername
- updatedByUserId
- updatedByUsername 
is not the part of the story, can you revert it? (it's possible that this field would never be used for searching purposes

pfilippov-epam (approved):

MikhailFokanov (approved):
Looks good.

PR #165, opened 2021-12-02
MSEARCH-208 Authority record search - filters/facets

Shans-Kaluhin:
*Purpose/Overview:*
 The purpose of this story is to create facets/filters that will provide a quick way to narrow down search results.

*Requirements/Scope:*

Following fields provide filter and faceted search:
 * *Personal name* based on personalName and sftPersonalName a
 * *Corporate name* based on corporateName and stfCorporateName properties
 * *Conference name* based on meetingName and stfMeetingName properties
 * *Geographic name* based on geographicName and stfGeographicName properties
 * *Uniform title* based on uniformTitle and stfUniformTitle properties
 * *Topical* based on topicalTerm and stfTopicalTerm properties
 * *Genre* based on genreTerm and stfGenreTerm properties
* *Other*  that would include all records that do not match above criteria

*Acceptance criteria:*
 * Query returns accurate hit count
 * Query returns only instances that fulfill selected facet
 * Filter can be combined and represent logical conjunction
 * If the value for suppress from discovery or staff suppress is set to undefined treat it as false
 * Filters are part of the cql query and the fields which are filtered should be configured in mappings - *TODO:* do we need that specified for authority records as well?

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -53,6 +53,11 @@
         },
         {
           "methods": [ "GET" ],
+          "pathPattern": "/search/authorities/facets",
+          "permissionsRequired": [ "search.authorities.facets.collection.get" ]
+        },
+        {
+          "methods": [ "GET" ],
           "pathPattern": "/search/instances/facets",
           "permissionsRequired": [ "search.instances.facets.collection.get" ]
         },
@@ -214,6 +219,11 @@
       "permissionName": "search.index.inventory.reindex.post",
       "displayName": "Search - starts inventory reindex operation",
       "description": "Starts inventory reindex operation"
+    },
+    {
+      "permissionName": "search.authorities.facets.collection.get",
+      "displayName": "Search - returns facets for a query for given filter options",
+      "description": "Returns facets for a query for given filter options"
     },
     {
       "permissionName": "search.instances.facets.collection.get",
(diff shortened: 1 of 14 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -51,6 +51,11 @@
...
           "permissionsRequired": [ "search.authorities.collection.get" ]
         },
+        {
+          "methods": [ "GET" ],
+          "pathPattern": "/search/authorities/facets",
+          "permissionsRequired": [ "search.authorities.facets.collection.get" ]
+        },
         {
           "methods": [ "GET" ],
           "pathPattern": "/search/instances/facets",
```
Shans-Kaluhin, line 61: Should we generalize this endpoint and passing parameter "recodType". 
Instead of creating two absolutely identical endpoints

src/test/java/org/folio/search/controller/AuthorityControllerTest.java
```diff
@@ -28,8 +28,10 @@
 @WebMvcTest(AuthorityController.class)
 class AuthorityControllerTest {
 
-  @MockBean private SearchService searchService;
-  @Autowired private MockMvc mockMvc;
+  @MockBean
+  private SearchService searchService;
+  @Autowired
+  private MockMvc mockMvc;
```
pfilippov-epam, line 34: please, revert formatting for this class

pfilippov-epam (approved):
Looks good except one thing

OleksiiKuzminov (approved):

PR #169, opened 2021-12-06
Fix potential NPE issue in sonarcloud

pfilippov-epam:
This issues is raised by sonarcloud, however it's impossible to retrieve null value for total hits according to Elasticsearch response contract


MikhailFokanov (approved):
Looks good.

PR #172, opened 2021-12-07
MSEARCH-238 Keyword search option

Shans-Kaluhin:
### Purpose/Overview:

The purpose of this story is to support Authority record search by authority by keyword

### Assumptions:
[MSEARCH-238](https://issues.folio.org/browse/MSEARCH-238)

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,41 +348,42 @@ if it is defined but doesn't match.
 
 ### Authority search options
 
-| Option                  |   Type    | Example                                      | Description                                                         |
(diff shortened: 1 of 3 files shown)


pfilippov-epam (changes requested):
Looks like that identifier field is not listed in the keyword search, can you fix it?

pfilippov-epam (approved):

MikhailFokanov (approved):
Looks good.

PR #173, opened 2021-12-09
MSEARCH-240 Improvements for retrieving instance ids

Shans-Kaluhin:
Purpose/Overview:
Currently instance ids are being returned as json stream, which contains a lot of redundant json events (like START_OBJECT and END_OBJECT), which decreases performance upon retrieving large sets of UUIDs. When dealing with large sets, it is better to return UUIDs as text values, separated by line breaks.

Requirements/Scope:
Improve /search/instances/ids endpoint to switch between json and text streams based on Accept header value (application/json or application/text). In case when it is not applicable, introduce new endpoint to retrieve instance ids as text stream.

Acceptance criteria:
UUIDs should be streamed as text values, separated by line breaks.


OleksiiKuzminov (approved):

src/main/java/org/folio/search/service/ResourceIdService.java
```diff
@@ -28,7 +30,22 @@
...
-   * Returns resource ids for passed cql query.
+   * Returns resource ids for passed cql query in text type.
+   *
+   * @param request      resource ids request as {@link CqlResourceIdsRequest} object
+   * @param outputStream output stream where text will be written in.
+   */
+  public void streamResourceIdsInTextType(CqlResourceIdsRequest request, OutputStream outputStream) {
+    try (var writer = new OutputStreamWriter(outputStream, UTF_8)) {
+      streamResourceIds(request, ids -> writeRecordIdsToOutputStream(ids, writer));
+    } catch (IOException e) {
```
pfilippov-epam, line 41: these new catch conditions are not covered with unit tests, can you fix it?

src/main/java/org/folio/search/service/ResourceIdService.java
```diff
@@ -82,4 +99,20 @@ private static void writeRecordIdsToOutputStream(List<String> recordIds, JsonGen
...
+    if (CollectionUtils.isEmpty(recordIds)) {
+      return;
+    }
+
+    try {
+      for (var recordId : recordIds) {
+        outputStreamWriter.write(recordId + '\n');
+      }
+      outputStreamWriter.flush();
+    } catch (IOException e) {
```
pfilippov-epam, line 113: this must be covered with unit test too

pfilippov-epam (commented):
Looks good, except few things related to code coverage

src/test/java/org/folio/search/controller/InstanceControllerTest.java
```diff
@@ -172,7 +175,31 @@ void getInstanceIds_positive() throws Exception {
...
+
+  @Test
+  void getInstanceIdsTextType_positive() throws Exception {
+    var cqlQuery = "id=*";
+    var instanceId = randomId();
+    var request = CqlResourceIdsRequest.of(INSTANCE_RESOURCE, TENANT_ID, cqlQuery, INSTANCE_ID_PATH);
+
+    doAnswer(inv -> {
+      var out = (OutputStream) inv.getArgument(1);
+      out.write(OBJECT_MAPPER.writeValueAsBytes(instanceId));
```
pfilippov-epam, line 191: as I understand correctly, this writer will provide following response
```
["instanceId1", "instanceId2", "instanceId3"...]
```

pfilippov-epam (approved):
Looks good, thanks!

PR #174, opened 2021-12-09
[WIP] MSEARCH-166 Implement system user from security store

pfilippov-epam:
### Purpose/Overview:

Tenant and module users (aka system users) should be created on adding of the new tenant and should be hidden from the users' app to reduce the risk of being accidentally changed/removed/etc.

### Requirements/Scope:

- The user shouldn't be created on postTenant, but the existing user should be used
- Name of the user should be the same as the module name
- Password should be taken from store variable (the same as it is implemented in edge-common). The name of the variable should be `$environment_$tenantId_$moduleName`. The envPrefix is the same, that is used for Kafka.
- The credentials for AWS Param store will be provided via environment variables (namely `client_id`, `client_secret`, `region`)
- If the parameters are not provided the module start should fail
- The required permissions must be updated for the system user during application deployment

### Approach
- Remove old functionality related to system user creation on post tenant operation
- Add dependency `edge-api-utils:1.1.0` which provides access to secure stores, token cache and their configurations
- Make all existing variables overridable from env configuration


descriptors/ModuleDescriptor-template.json
```diff
@@ -252,6 +252,14 @@
       "description": "Removes feature configuration"
     }
   ],
+  "user": {
```
wafschneider, line 255: I don't think this is legal within the module descriptor schema, which has `"additionalProperties": false`. It might be able to be embedded within the `metadata` property of the module descriptor. Have you tried posting this module descriptor to Okapi? @adamdickmeiss may have more insight.

wafschneider:
I do not find any documentation for operators in this pull request. Please add some documentation to the README "Deployment" section so folks trying to deploy this have a chance of succeeding.

descriptors/ModuleDescriptor-template.json
adamdickmeiss: Right. Not legal

adamdickmeiss:
> I do not find any documentation for operators in this pull request. Please add some documentation to the README "Deployment" section so folks trying to deploy this have a chance of succeeding.



> I do not find any documentation for operators in this pull request. Please add some documentation to the README "Deployment" section so folks trying to deploy this have a chance of succeeding.

I think thus means mod-search becomes an edge module. It would have to be enabled separately as it requires system users to be populated before it can operate. A very inelegant solution for a backend to backend module communication which should just have a token (no user, no password)

descriptors/ModuleDescriptor-template.json
MikhailFokanov-ebsco: @wafschneider this is addressed in https://github.com/folio-org/okapi/pull/1168

PR #176, opened 2021-12-13
MSEARCH-177 Prevent resource skipping during resource indexing

pfilippov-epam:
### Purpose
Current implementation ignores the whole resource(inventory instance) if the reference data endpoint returns an error response (BAD_REQUEST, FORBIDDEN).

### Requirements/Scope
* An resource must be indexed to Elasticsearch with logging event when reference data endpoint returns an error.

### Acceptance criteria:
* Resources must be uploaded to Elasticsearch in spite of reference data available or not
* Integration and unit tests are added to prove functionality

### Approach
- Refactored retrieving reference data using single feign client
- Add logging if reference data is not loaded correctly
- Add unit and spring-based test to verify that caching is working correctly


OleksiiKuzminov (approved):

MikhailFokanov (commented):
Looks good.

PR #183, opened 2021-12-24
MSEARCH-253 Implement call-number browsing

pfilippov-epam:
### Purpose
Users should be able to browse call numbers based on a provided whole or partial call number.

### Requirements

- Searched call number will be considered as an anchor point for the search results
- The number of returned records should be configurable
- The number of preceding matches should be configurable
- The number of succeeding matches should be configurable
- Each call number should be followed by the number of instances matching the search
- If there is more than one record matching the searched call number, the title should not be included
- Matching record can be marked so that UI can display it correctly
- If a matching record is not found then its placeholder will need to be marked as well
- The browse will be based on the effective shelf order value of the item record

### Approach

* Implement call-number browsing using normalization of item.effectiveShelvingOrder field value. The value is normalized according to the following rules:
    * Number format is 0000_000000_00000_00000 as double
    * Four first numbers are responsible for general letters: `A=100, AA=2, B=28, Z=676, ZZ=704`
    * Next six numbers are responsible for classification value: `11=110000, 11.1=110001, 512345.6=512345`. Values longer than 6 characters are truncated
    * Next five numbers are responsible for first cutter: `A1=1100, A111=01111, Z999=26999`. The first two numbers are responsible for the letter, the next three - for the numeric part. All numeric values longer than 3 characters are truncated.
    * The last five numbers are responsible for the second cutter. It's processed with the same approach as the first cutter. If it's missing - it will be populated with five zeroes: `00000`.
* Implement range-query based approach to retrieve related instances around using script sorting based on the difference between calculated call-number value and anchor
* Implement CallNumberBrowseService with required CQL query validations and necessary logic to perform call-number browsing using existing algorithms
* Add integration and unit test to prove functionality


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

PR #191, opened 2022-01-19
MSEARCH-254 Implement subject browsing

pfilippov-epam:
### Purpose
Users should be able to browse call numbers based on instance records' subjects. 

### Approach
* Implement subject browsing endpoint and corresponding service
* Add unit and integration tests
* Update module descriptors with a new endpoint


MikhailFokanov-ebsco (approved):
Looks good. As Pavel tested the size of the subjects index, and it is ~10% of instances index.

PR #193, opened 2022-01-19
MSEARCH-272 Include Item identifier in Identifier (all) search

OleksiiKuzminov:
### Purpose
The existing implementation of Item Identifiers (all) includes searches for items' HRIDs, Former Ids, and Accession numbers.   However, the libraries might use Item Identifiers as well, for example for a secondary remote storage identifier.

### Approach
item and holding UUID added to the corresponding processors.


pfilippov-epam (approved):

MikhailFokanov (approved):
Looks good.

PR #198, opened 2022-01-25
MSEARCH-256 Implement authority record browsing

pfilippov-epam:
### Purpose
Users should be able to browse the following authority fields:
- personalName
- sftPersonalName
- personalNameTitle
- sftPersonalNameTitle
- corporateName
- sftCorporateName
- corporateNameTitle
- sftCorporateNameTitle
- meetingName
- sftmeetingName
- meetingNameTitle
- sftmeetingNameTitle
- geographicName
- sftgeographicName
- uniformTitle
- sftuniformTitle
- topicalTerm
- sfttopicalTerm
- genreTerm
- sftgenreTerm

### Requirements/Scope
- Searched heading or reference (also known as sft) will be considered as an anchor point for the search results
- The number of returned records when you display or hit Previous or Next pagination should be configurable (default 100)
- The number of preceding matches of the anchor result should be configurable (default 5)
- Succeeding results will be calculated
- The results should be ordered in the alphabetical order
- The response should contain headings, references
- Search Match record can be marked so that UI can display it correctly
= IF no search match is found then display an empty authority object as a placeholder for UI

#### Approach
- Reuse Subject browsing approach but without adding an additional index to perform the browsing because there is no need to calculate counts of unique authorities
- Add new endpoint and related settings to the module descriptor
- Add unit and integration tests to prove functionality


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

PR #200, opened 2022-01-27
MSEARCH-186 Add isBoundWith field to instance

pfilippov-epam:
### Purpose
`mod-search` must publish the isBoundWith attribute of instances where that attribute is true

### Approach
- Add field `isBoundWith` to the response model and resource description
- Update integration tests to verify that the new field is included in the response


MikhailFokanov (approved):
Also looks good.

OleksiiKuzminov (approved):

PR #201, opened 2022-01-27
MSEARCH-254 Fix the reindex parameter to clear indices for secondary resources

pfilippov-epam:
### Purpose
The reindex parameter must clear indices for the target resource and all of the secondary resources too

### Approach
- Add to resource description parameter - `parent` (which must be the resource name of the parent resource description)
- Add unit tests to check that the new clear indices approach on `recreateIndex` request is working correctly


OleksiiKuzminov (approved):
Looks good

MikhailFokanov (approved):
Looks good.

PR #202, opened 2022-01-28
MSEARCH-283 Fix call number exists query

pfilippov-epam:
### Purpose
Query ```callNumber = ""``` must return all results containing generated callNumber.

### Approach
- Fix an issue when `CallNumberBrowseSearchTermProcessor` returns 0 for empty value
- Add unit and integration test to verify the new approach


MikhailFokanov (approved):
Looks good.

PR #204, opened 2022-02-09
MSEARCH-276 Support delete events for instance subjects

pfilippov-epam:
### Purpose
Current implementation must be extended to support delete operation on instance subject index to delete subject that does not have corresponding instances after delete events from Kafka.

### Approach
- Move indexing logic to the ResourceRepository
- Implement custom repositories for delete events on instance_subject
- Implement custom event pre-processing and indexing
- Add subject count query
- Implement delete events for subjects
- Fix Spring Elasticsearch deprecated properties
- Refactor SearchDocumentBody to store related resource event


MikhailFokanov (approved):
Looks excellent.

OleksiiKuzminov (approved):

PR #209, opened 2022-02-11
MSEARCH-276 Fix update operation on instance

pfilippov-epam:
### Purpose
This pull-request fixing issue was noticed in the rancher environment when update events have been ignored

### Approach
- Refactor the ResourceFetchService to pass further the old body within the new resource event
- Fix issue when ResourceService ignores UPDATE events
- Add integration test to verify that update operation works correctly


MikhailFokanov (approved):
Looks good.

PR #219, opened 2022-03-16
MSEARCH-301 Implement browsing by Dewey Decimal and Other schema numbers

pfilippov-epam:
### Purpose
Users should be able to browse call numbers based on a provided whole or partial call number. 

Requirements/Scope:

- Searched call number will be considered as an anchor point for the search results
- The number of returned records should be configurable
- The number of preceding matches should be configurable
- The number of succeeding matches should be configurable
- Each call number should be followed by the number of instances matching the search
- If there are more than one record matching the searched call number, the title should not be included
- Matching record can be marked so that UI can display it correctly
- If matching record is not found then its placeholder will need to be marked as well (UI?)
- The browse will be based on the effective shelf order value of the item record

### Approach
- Current approach supports any call-number value in field `items.effectiveShelvingOrder` and can be applied to LC, Dewey Decimal and Other schema numbers
- Implement call-number browsing based on the range queries (boundaries are provided by the numeric representation of real shelf-key (used only 10 first characters, limit by the long value capacity)
- Implement algorithm to convert String call-number into long value for range queries
- Reimplement call-number browsing approach to use range queries to increase performance and new script sorting to arrange search results
- Implement of populating and cleaning-up the incoming search results (new items will be populated if they have call-number between selected ranges (1st and last result), ambiguous results are removed from the browsing page
- Search aliases can be applied in the range queries
- Refactor and split the CallNumberBrowseService into smaller parts to make it maintainable and easier to test and read

### Learning
- Range queries can help to optimize search request in Elasticsearch by reducing the number of resources to scan
- New script sorting is developed to implement alphabetically correct search results order
- Here are some examples of the Call Number Browse:
  - University of Chicago: https://catalog.lib.uchicago.edu/vufind/Alphabrowse/Home?source=lcc&from=A&page=0
  - Michigan State University: https://magic.msu.edu/search/c?SEARCH=pr&sortdropdown=-&searchscope=39
  - Center for Research Libraries : http://catalog.crl.edu/search/e?SEARCH=L&searchscope=2&x=0&y=0
  - Cornell University: https://newcatalog.library.cornell.edu/browse?authq=a&start=0&browse_type=Call-Number


MikhailFokanov (approved):
Tremendously good.

OleksiiKuzminov (approved):

PR #220, opened 2022-03-17
MSEARCH-297 SearchAliases not working for fields with "searchTermProcessor"

Shans-Kaluhin:
### Purpose
SearchAliases not working for fields with "searchTermProcessor"

### Approach
If we use searchAliases name, it use empty `searchTermProcessor`,
even if field has `searchTermProcessor`

To fix that, I change name to original to find right `searchTermProcessor`

### Learning
[MSEARCH-297](https://issues.folio.org/browse/MSEARCH-297)


psmagin (commented):
Could you please cover the change with tests?

Shans-Kaluhin:
> Could you please cover the change with tests?

I do, but we don't have fields with searchAliases and searchTermProcessor. (Bad idea to add new field just for test)
I wrote and run tests when explored bug

psmagin:
> > Could you please cover the change with tests?
> 
> I do, but we don't have fields with searchAliases and searchTermProcessor. (Bad idea to add new field just for test) I wrote and run tests when explored bug

I'm sure that in the `CqlTermQueryConverterTest` test you can add some fields to test this change.

Shans-Kaluhin:
@psmagin Good idea! 
Also I find that for holdings and item fields we add searchAliases and separate fields with dots.
So now we have fields with searchAliases and searchTermProcessor

psmagin (approved):

PR #222, opened 2022-03-19
MSEARCH-282 Implement normalized search option for OCLC identifiers

psmagin:
### Purpose
Implement normalized search option for OCLC identifiers

### Approach
* add new searchFields `oclc`
* implement field processor and search term processor
* update README


src/main/java/org/folio/search/service/setter/instance/OclcProcessor.java
```diff
@@ -0,0 +1,74 @@
...
+      .map(this::normalizeOclc)
+      .filter(Objects::nonNull)
+      .collect(toCollection(LinkedHashSet::new));
+  }
+
+  /**
+   * Returns normalized isbn value.
+   *
+   * @param value value to process as {@link String}
+   * @return normalized isbn value
```
Shans-Kaluhin, line 49:
```suggestion
   * Returns normalized oclc value.
   *
   * @param value value to process as {@link String}
   * @return normalized oclc value
```

psmagin: fixed

PR #226, opened 2022-03-24
MSEARCH-306 Gracefully handle 500 error after the search reaches index.max_result_window.

DieHardJava:
Gracefully handle 500 error after the search reaches index.max_result_window.
### Purpose/Overview:

Exception with status code 400 should be thrown, when user reaches last page of result list with 10k records instead of status code 500
 
### Requirements/Scope:
 - max validation should be put on top of offset parameter in searchInstances method
 - bad request should be thrown when offset reaches 10.000, also when sum of offset and limit exceeds 10000

### Approach
 - set maximum value to 9999 in offset parameter in mod-search.yml file
 - set validation if sum of the offset and limit exceeds 10000 throws RequestValidationException
 - write unit test to make sure it works correctly


src/test/java/org/folio/search/controller/InstanceControllerTest.java
```diff
@@ -74,6 +74,25 @@ void search_positive() throws Exception {
...
+  void search_offset_limit_10k() throws Exception {
+
+    var cqlQuery = "title all \"test-query\"";
+
+    when(searchService.search(searchServiceRequest(Instance.class, cqlQuery))).thenReturn(searchResult());
+
+    var requestBuilder = get("/search/instances")
+      .queryParam("query", cqlQuery)
+      .queryParam("limit", "100")
+      .queryParam("offset", "10000")
```
pfilippov-epam, line 87: try this case 
```
limit = 500 (allowed)
offset = 9600
```

DieHardJava: at the moment max allowed value for limit is 500, I need to just change values in test to these?

DieHardJava: in this case test is failed

pfilippov-epam (approved):

OleksiiKuzminov (approved):

PR #230, opened 2022-03-30
MSEARCH-315 Implement cursor parameters for browsing

pfilippov-epam:
### Purpose
Next\Prev buttons must be disabled on UI if there is no data in forward\backward directions, mod-search must mark the first and last pages of browse results.

### Approach
- It's impossible to mark the first and the last page because browsing operates with a term - `anchor` which is actually cursor, and `prev`/`next` values can be helpful to implement it.
- Add cursor parameters for browsing for backward and forward directions:
  - **around/around_including**: the `prev` and `next` values will be filled with values for forward and backward queries if there are any records, one of them can be null or undefined if browsing in the following direction will produce zero results.
  - **forward/foward_including**: only `next` value is specified, it can be null if there are no records ahead
  - **backward/bacward_including**: only `prev` value is specified, it can be null if there are no records below


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

OleksiiKuzminov (approved):

PR #235, opened 2022-04-05
MSEARCH-334 Fix the invalid search results for call-number with 2+ spaces

pfilippov-epam:
### Purpose
Call-numbers with double spaces after leading letter prevents from browsing in backward direction.

### Approach
- Fix the `CallNumberProcessor` to correctly process values with double-spaces


OleksiiKuzminov (approved):

MikhailFokanov (approved):
Looks good.

PR #238, opened 2022-04-07
MSEARCH-324 Use match all query for 'keyword=*'

pfilippov-epam:
### Purpose
Assuming that the instance title is mandatory for each resource - the wildcard query can be replaced with match_all query, which will increase the performance when the user will want to check the number of results.

### Approach
- Use elasticsearch `match_all` query for input like `keyword = *`
- Add integration and unit tests to validate it


MikhailFokanov (approved):
Thanks a lot. It is what I asked for from the beginning.

OleksiiKuzminov (approved):

PR #243, opened 2022-04-21
MSEARCH-339 Call-Number browse: invalid search result call-number with ignored characters

pfilippov-epam:
### Purpose
Following call-numbers prevents backward direction for call-number browsing:
```
F-1,452
F-1,490
F-1,720
```
This is caused by the fact that implemented approach to transform string value to number ignores the comma and hyphen characters. Some of call-numbers can be normalized or ignored to provide better browsing experience without issues

### Approach
- All unsupported characters are ignored for string and numeric values and now they are consistent
- New field is added for browsing, previous marked as source
- Nested item sorting is removed (became redundant)


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

PR #246, opened 2022-04-28
MSEARCH-342 Add default prev/next values for browsing forward/backward

pfilippov-epam:
### Purpose
In current implementation - for forward the previous value is null, for backward - next value is null. It can be improved by adding this value from the result list, assuming that these queries can be reached after initial request by browsing around.

### Approach
- Previous and next values are calculated on the existing dataset and they are provides more comfortable way to browse the catalog
- Unit and integration test are updated with the new field values


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

PR #247, opened 2022-04-29
MSEARCH-344 Implement response fields per feature/endpoint

pfilippov-epam:
### Purpose
Returning required fields per feature/endpoint could allow in future customize response body to minify its size

### Approach
- Replace boolean value for `showInResponse` with array of predefined response groups
- Update the related documentation and tests


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

PR #265, opened 2022-06-07
MSEARCH-365 Add support of 'string' modifier for '=='

psmagin:
### Purpose
Add support of 'string' modifier for '=='

### Approach 
If `==` operator has modifier `string` and used for full text field - then use term search instead of phrase search


src/main/java/org/folio/search/cql/builders/ExactTermQueryBuilder.java
```diff
@@ -20,14 +21,16 @@ public class ExactTermQueryBuilder extends FulltextQueryBuilder {
 
   public static final String SCRIPT_TEMPLATE = "doc['%s'].size() == 0";
 
+  private static final String STRING_MODIFIER = "string";
```
viacheslavkol, line 24: Maybe it would be better to move this into enum?

psmagin: I thought about this. But until there is only one modifier option I think it's okay to have just a constant.

Shans-Kaluhin (approved):

viacheslavkol (approved):

PR #268, opened 2022-06-22
MSEARCH-383 Change contributor events producing to ignore contributorTypeText

psmagin:
### Purpose
Fix bug when existed contributor is not returned at Browse contributors result list when user filled field `contributorTypeText`.

### Approach
Ignore `contributorTypeText` when forming contributor events. If taking this field into an account then system formes one update and one delete event, and contributor is removed from index.

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,9 +505,9 @@ if it is defined but doesn't match.
 
 ### Contributors search options
 
-| Option              | Type | Example                         | Description                                     |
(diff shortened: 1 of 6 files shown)


Shans-Kaluhin (approved):
Could you also please update contributors facet 
`contributorTypeId` -> `contributorNameTypeId` in
Readme file

viacheslavkol (approved):

PR #277, opened 2022-07-19
MSEARCH-393 Fix date query validation to support ISO date/time formats

psmagin:
## Purpose
Fix query validation when param is date-time

## Approach
Make date query validation to support ISO date/time formats

### Changes checklist
- [ ] Database schema changes


src/main/java/org/folio/search/cql/CqlTermQueryConverter.java
```diff
@@ -143,11 +152,24 @@ private static Map<String, TermQueryBuilder> getTermQueryProvidersAsMap(List<Ter
...
+    boolean isValidDate = false;
+    for (DateTimeFormatter dateFormat : SUPPORTED_DATE_FORMATS) {
+      try {
+        dateFormat.parse(value);
+        isValidDate = true;
+      } catch (Exception ignored) {
+        // do nothing
+      }
+    }
+    return isValidDate;
```
oliinyko, line 170: I assume when the first parses is successfully applied then there is no need to check the rest of them
 ```suggestion
    for (DateTimeFormatter dateFormat : SUPPORTED_DATE_FORMATS) {
      try {
        dateFormat.parse(value);
        return true;
      } catch (Exception ignored) {
        // do nothing
      }
    }
  return false;
```

oliinyko (approved):

PR #278, opened 2022-07-25
MSEARCH-392 Stream ids job, relation doesn't exist while using multi tenant

Shans-Kaluhin:
## Purpose
The errors like this appears when additional tenant was added:
`ERROR: relation "resource_ids_job" does not exist. (at ResourceIdService:110)`

It looks like the repository can't find the correct schema because the async job is running in transaction 

## Approach
- Starting async job execution in scope of current tenant
- Separate @Async and @Transaction annotations (They can't used together)
- Add Content-Type for response. Without this header karate can't display the body response [KARATE issue](https://github.com/karatelabs/karate/issues/1024#issuecomment-1195374833)

### Learning
https://issues.folio.org/browse/MSEARCH-392

### DIFF
--- a/src/main/java/org/folio/search/service/ResourceIdsJobService.java
+++ b/src/main/java/org/folio/search/service/ResourceIdsJobService.java
@@ -26,7 +26,6 @@ public class ResourceIdsJobService {
     return resourceIdsJobMapper.convert(jobRepository.getById(id));
   }
 
-  @Transactional
   public ResourceIdsJob createStreamJob(ResourceIdsJob job, String tenantId) {
     var tableName = RandomStringUtils
       .random(32, 0, 0, true, false, null, new SecureRandom())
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/search/service/ResourceIdsJobService.java
```diff
@@ -28,15 +29,21 @@ public ResourceIdsJob getJobById(String id) {
 
   @Transactional
```
psmagin, line 30: I'm thinking about if this annotation is needed there because all db-related work is done in `resourceIdService.streamResourceIdsForJob` which is already transactional.

Shans-Kaluhin: Yes, here we just save first job. Looks like this annotation is optional here

psmagin (approved):

PR #284, opened 2022-08-10
MSEARCH-401 Call-number browsing fails when value match shelfKey pattern

Shans-Kaluhin:
## Purpose
Call-number browsing fails when value match shelfKey pattern

## Approach
- Remove pattern check. So shelfKey will be generated from call-number even if call-number matches shelfKey pattern.
- Remove `CallNumberTermConverter` and pass anchor directly. (Because anchor in context is always is a shelfKey)
- Include fullCallNumber to empty anchor

### Learning
https://issues.folio.org/browse/MSEARCH-401
https://falcon.ci.folio.org/inventory


src/main/java/org/folio/search/cql/EffectiveShelvingOrderTermProcessor.java
```diff
@@ -16,21 +11,12 @@
...
   public String getSearchTerm(String inputTerm) {
-    if (StringUtils.isEmpty(inputTerm) || anyMatch(PATTERNS, pattern -> pattern.matcher(inputTerm).matches())) {
-      return toRootUpperCase(inputTerm);
-    }
-
     return getValidShelfKey(new LCCallNumber(inputTerm))
       .or(() -> getValidShelfKey(new DeweyCallNumber(inputTerm)))
-      .orElse(normalizeValue(inputTerm));
+      .orElse(normalizeValue(inputTerm))
+      .trim();
```
Shans-Kaluhin, line 19: The same trimming as in mod-inventory-storage
image

src/main/java/org/folio/search/service/browse/CallNumberBrowseQueryProvider.java
```diff
@@ -65,7 +66,7 @@ public SearchSourceBuilder get(BrowseRequest request, BrowseContext ctx, boolean
 
   private QueryBuilder getQuery(BrowseContext ctx, String tenantId, int size, boolean isBrowsingForward) {
     var anchor = ctx.getAnchor();
-    var callNumberAsLong = callNumberTermConverter.convert(anchor);
+    var callNumberAsLong = callNumberProcessor.getCallNumberAsLong(anchor);
```
Shans-Kaluhin, line 69: the anchor from context is always a shelfKey.
So it's no need to generate from it shelfKey again.

psmagin (approved):

oliinyko (approved):

PR #289, opened 2022-08-22
MSEARCH-415: Option for Elasticsearch integration test

julianladisch:
In #cc-elasticsearch-opensearch-discussion Slack channel people asked whether mod-search can still run with an Elasticsearch server.

Provide an option to switch the search engine from OpenSearch to Elasticsearch when running the integration test phase of the build.

Implementation details:

OpenSearch and Elasticsearch consider an undefined property, a null property and an empty list property the same. Elasticsearch 8.1 changed the empty list to a null value in same cases. This is a compatible change. To make the integration tests work the swagger schema files for instance, holding and item have been changed to always provide an empty list to compensate for the change; this is why some array fields have been made required. The changes in the schema files need changes in ResourceFetchServiceTest.java.

There was no null/empty list change in OpenSearch yet, only in Elasticsearch.

### DIFF
--- a/src/main/resources/swagger.api/schemas/holding.json
+++ b/src/main/resources/swagger.api/schemas/holding.json
@@ -82,5 +82,6 @@
     "metadata": {
       "$ref": "metadata.json"
     }
-  }
+  },
+  "required": ["electronicAccess", "notes"]
 }
--- a/.github/workflows/elasticsearch.yml
+++ b/.github/workflows/elasticsearch.yml
@@ -1,0 +1,17 @@
+name: elasticsearch
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
+        java-version: '11'
+        cache: 'maven'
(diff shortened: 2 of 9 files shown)


src/main/resources/swagger.api/schemas/holding.json
```diff
@@ -82,5 +82,6 @@
     "metadata": {
       "$ref": "metadata.json"
     }
-  }
+  },
+  "required": ["electronicAccess", "notes"]
```
adamdickmeiss, line 86: More strict schema? what has it got to do with this PR?

julianladisch: This is explained in the "Implementation details" section of the pull request description.

julianladisch: If you want the client that calls mod-search to make the normalization I can remove the normalization in mod-search, and implement a hamcrest matcher that ignores these differences in the unit tests.

julianladisch:
@psmagin As described in the PR description Elasticsearch 8.1 changed the empty list to an undefined property in some same cases. The "required" in the instance, holding and item schema enforces the empty list. If you remove the "required" you will see that Elasticsearch 8.1 no longer produces an empty array, the property is undefined. This fails many unit tests that expect the empty array.

julianladisch:
This PR is split into two:
* Integration test option (MSEARCH-415) #291
* Upgrade from Elasticsearch 8.0.1 to 8.3.3.

PR #292, opened 2022-08-29
MSEARCH-414 Browse call-numbers. Search failed when record have enumeration

Shans-Kaluhin:
## Purpose
Fix searching by shelving order.
Fix browsing prev/next by shelving order.

## Approach
- Remove termProcessor for `itemEffectiveShelvingOrder`
- Return to using shelving order in prev/next
- Add sub-field for browsing.
  - Firstly we will browse around by call-number (`callNumber>="RR 720" or callNumber<"RR 720"`)
  - Then take prev/next shelving order and browse in one direction by shelving order 
    (`itemEffectiveShelvingOrder>"RR 3720"`)

### DIFF
--- a/src/main/resources/model/instance.json
+++ b/src/main/resources/model/instance.json
@@ -610,8 +610,7 @@
     "itemEffectiveShelvingOrder": {
       "type": "search",
       "index": "keyword",
-      "searchAliases": [ "callNumber", "items.effectiveShelvingOrder" ],
-      "searchTermProcessor": "effectiveShelvingOrderTermProcessor",
+      "searchAliases": [ "items.effectiveShelvingOrder" ],
       "processor": "itemEffectiveShelvingOrderProcessor"
     },
     "allInstances": {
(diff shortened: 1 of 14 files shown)


src/main/resources/model/instance.json
```diff
@@ -610,8 +610,7 @@
     "itemEffectiveShelvingOrder": {
       "type": "search",
       "index": "keyword",
-      "searchAliases": [ "callNumber", "items.effectiveShelvingOrder" ],
-      "searchTermProcessor": "effectiveShelvingOrderTermProcessor",
+      "searchAliases": [ "items.effectiveShelvingOrder" ],
```
viacheslavkol, line 613: Shouldn't it be 'item.effectiveShelvingOrder' ?

Shans-Kaluhin: Yes you're right. This searchAlias using for backward compatibility.
To have ability to search by "item" we using `ItemSearchFieldModifier` class
From tests you can see that searching by `item.effectiveShelvingOrder` works:
image

viacheslavkol: Ok, got it, thank you

viacheslavkol (approved):

oliinyko (approved):

psmagin (approved):

PR #294, opened 2022-08-31
MSEARCH-416 Similar call numbers from different items of one "Instance"

Shans-Kaluhin:
## Purpose
Similar call numbers from different items of one "Instance" record don't display at the browse result list one by one.

## Approach
- Enable `BROWSE_CN_INTERMEDIATE_VALUES_ENABLED` env. variable 
- Add `BROWSE_CN_INTERMEDIATE_REMOVE_DUPLICATES` env. variable

### Learning
https://issues.folio.org/browse/MSEARCH-416


src/test/java/org/folio/search/controller/BrowseCallNumberIT.java
```diff
@@ -170,35 +170,36 @@ private static Stream<Arguments> callNumberBrowsingDataProvider() {
           cnBrowseItem(instance("instance #22"), "DA 3890 A2 B76 42002")
         ))),
 
+      //TODO: fix this test
       // checks order of closely placed call-numbers
```
psmagin, line 174: Why not to fix it now?

Shans-Kaluhin: Fixed ✔️

oliinyko (approved):

viacheslavkol (approved):

PR #299, opened 2022-09-08
MSEARCH-427: Add integration test option for Elasticsearch 8

julianladisch:
## Purpose

Elasticsearch 7.17.* is LTS with EOL date 2023-08-01: https://www.elastic.co/support/eol#elasticsearch

However, people on #cc-elasticsearch-opensearch-discussion Slack channel asked for the compatibility with latest Elasticsearch version.

## Approach

Add an additional Dockerfile for the latest Elasticsearch 8 version (8.4.1) one can manually switch to before running the mod-search integration tests. The README.md is updated to mention this new option.

OpenSearch (all versions) and Elasticsearch before 8.1.0 returns an empty list when a list join yields no element. Elasticsearch from 8.1.0 on returns null when a list join yields no element.

This fails a few tests:
* BrowseCallNumberIT.browseByCalNumber_browseAroundWithEnabledIntermediateValues:111
* BrowseCallNumberIT.browseByCallNumber_browsingAroundWhenPrecedingRecordsCountIsSpecified:72
* BrowseCallNumberIT.browseByCallNumber_browsingAroundWithoutHighlightMatch:91
* BrowseCallNumberIT.browseByCallNumber_parameterized:61
* BrowseCallNumberOtherIT.browseByCallNumber_browsingAroundWithDisabledIntermediateValues:53
* BrowseCallNumberOtherIT.browseByCallNumber_browsingAroundWithEnabledIntermediateValues:78
* SearchInstanceIT.responseContainsAllInstanceProperties:136

To compensate for this change in Elasticsearch 8.1.0 mod-search's API schemas are changed: The affected list properties are declared as required, this automatically changes a null result into an empty list.

The returned JSON is the same as before.

The schemas are used for GET APIs only, therefore this schema change is a non-breaking change. And the JSON response is the same as before.

Quote from https://www.elastic.co/guide/en/elasticsearch/reference/current/null-value.html "A null value cannot be indexed or searched. When a field is set to null, (or an empty array or an array of null values) it is treated as though that field has no values."

### Changes checklist

- [ ] Database schema changes


src/main/resources/swagger.api/schemas/instance.json
```diff
@@ -194,6 +194,7 @@
         "$ref": "holding.json"
       }
     }
-  }
+  },
+  "required": ["electronicAccess", "notes", "items", "holdings"]
```
psmagin, line 198: Why only these fields? There are much more arrays in the schema?
What will happen when an incoming record doesn't have these fields?

julianladisch: Other fields are not a result of a join executed by Elasticsearch and therefore are not affected.

src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -105,7 +105,7 @@ void responseContainsOnlyBasicInstanceProperties() throws Exception {
     var response = doSearchByInstances(prepareQuery("id=={value}", getSemanticWebId()))
       .andExpect(jsonPath("totalRecords", is(1)))
       // make sure that no unexpected properties are present
-      .andExpect(jsonPath("instances[0].length()", is(8)));
+      .andExpect(jsonPath("instances[0].length()", is(11)));
```
marcjohnson-kint, line 108: What has caused this expectation to change?

Why would moving to Elastic Search 8 mean the expectations for a API response changes?

julianladisch: SearchInstanceIT doesn't use Elasticsearch but MockMvc that is initialized with an instance of a Instance.java: https://github.com/folio-org/mod-search/blob/e7cde64376c467a8fb62688f2c989c28dce5253b/src/test/java/org/folio/search/controller/SearchInstanceIT.java#L39-L42

This pull requests adds
```
"required": ["electronicAccess", "notes", "items", "holdings"]
```
to instance.json making these array properties an empty array by default in Instance.java. Before they were null/omitted by default.

"items" array is not affected because it is an array with two elements in this unit test.

marcjohnson-kint: Thanks. 

I take that mean that this is a side effect of changing the API definition to change the serialization to include empty arrays for these properties.

Is that correct?

julianladisch: Yes.

marcjohnson-kint: Thank you for clarifying. I cannot resolve this comment, please do so if you can.

docker/elasticsearch8/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM elasticsearch:8.4.1
+
+# Switch from /dev/random to /dev/urandom, otherwise the plugin installer needs many minutes to get
+# random numbers to verify the signature of the downloaded plugins.
+# https://github.com/elastic/elasticsearch/pull/84766
+# https://stackoverflow.com/questions/137212/how-to-deal-with-a-slow-securerandom-generator
+# https://security.stackexchange.com/questions/3936/is-a-rand-from-dev-urandom-secure-for-a-login-key
+USER root
+RUN sed -i 's!^securerandom\.source=.*!securerandom.source=file:/dev/./urandom!' /usr/share/elasticsearch/jdk/conf/security/java.security
+USER elasticsearch
```
marcjohnson-kint, line 10: Thank you for documenting the reasoning behind this step

psmagin (approved):

viacheslavkol (approved):

PR #304, opened 2022-10-04
MSEARCH-433 BE - MARC authority: Search Results/Browse List: Add a new column Number of titles

maryna-steshenko:
## Purpose
Add a new column "Number of titles" to Search Results/Browse List of MARC Authority and populate it with total number of bib records linked to this authority record.
[MSEARCH-433](https://issues.folio.org/browse/MSEARCH-433)

## Approach
- numberOfTitles field added to authority schema.
- NumberOfTitlesPostProcessor added to populate numberOfTitles field value.
- Integration tests in SearchAuthorityIT and BrowseAuthorityIT updated to check numberOfTitles field.

### Changes checklist
- [ ] Database schema changes


psmagin (changes requested):
1. Module descriptor should be updated
2. Add integration test in SearchAuthorityIT

viacheslavkol (changes requested):
Permission for API "/authorities/bulk/count" should be added to "modulePermissions" section of corresponding mod-search APIs in `ModuleDescriptor-template.json`

src/main/java/org/folio/search/service/setter/authority/NumberOfTitlesPostProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+public final class NumberOfTitlesPostProcessor implements PostProcessor {
+
+  private final EntitiesLinksClient entitiesLinksClient;
+
+  @Override
+  public void process(List<Object> res) {
+    if (res == null || res.isEmpty() || !(res.get(0) instanceof Authority)) {
+      return;
+    }
+
```
psmagin, line 25: I suggest making PostProcessor generic by type of response object. 
In `ElasticsearchDocumentConverter` you can create a map of processors with a key by class. By doing so, you can get all processors for the exact object. This will decrease processing time for objects that are not require some postprocessing

descriptors/ModuleDescriptor-template.json
```diff
@@ -215,6 +215,10 @@
     {
       "id": "instance-reindex",
       "version": "0.1"
+    },
+    {
+      "id": "instance-authority-links",
```
psmagin, line 220: You must add modulePermission to endpoint(s) that used this interface

src/main/java/org/folio/search/service/setter/authority/NumberOfTitlesPostProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+      .map(Authority::getId)
+      .map(UUID::fromString)
+      .collect(toList());
+
+    var numbersOfTitles = entitiesLinksClient.getLinksCount(
+      EntitiesLinksClient.UuidCollection.of(authorityIds)).getBody();
+
+    IntStream.range(0, authorizedAuthorities.size())
+      .forEach(i -> authorizedAuthorities.get(i).setNumberOfTitles(
+        getNumberOfTitles(numbersOfTitles, UUID.fromString(authorizedAuthorities.get(i).getId()))));
```
psmagin, line 40: It's not the best solution to use such stream. I think will be better to convert response to map with key=authId and value=count. Then you can easily find count for authority

psmagin (changes requested):

psmagin (approved):

PR #307, opened 2022-10-11
MSEARCH-439 Transferring items from one instance to another results in multiple matches when searching

viacheslavkol:
## Purpose
Remove item from old instance when moved

## Approach
- Divide event into two separate when "old" instanceId differs form "new" instanceId

### Changes checklist
- [ ] Database schema changes

### Learning
[MSEARCH-439](https://issues.folio.org/browse/MSEARCH-439)

### DIFF
--- a/src/main/java/org/folio/search/service/ResourceService.java
+++ b/src/main/java/org/folio/search/service/ResourceService.java
@@ -150,9 +150,9 @@ public class ResourceService {
   }
 
   /**
-   * There may be a case when some data is moved between instances
-   * In such case old and new fields of the event will have different instanceId
-   * This method will create 2 events out of 1 and erase 'old' field in an original event
+   * There may be a case when some data is moved between instances.
+   * In such case old and new fields of the event will have different instanceId.
+   * This method will create 2 events out of 1 and erase 'old' field in an original event.
    * */
   private List<ResourceEvent> extractEventsForDataMove(List<ResourceEvent> resourceEvents) {
     if (resourceEvents == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/search/service/ResourceService.java
```diff
@@ -142,6 +149,36 @@ private <T> List<T> getEventsThatCanBeIndexed(List<T> events, Function<T, String
...
+   * In such case old and new fields of the event will have different instanceId.
+   * This method will create 2 events out of 1 and erase 'old' field in an original event.
+   * */
+  private List<ResourceEvent> extractEventsForDataMove(List<ResourceEvent> resourceEvents) {
+    if (resourceEvents == null) {
+      return Collections.emptyList();
+    }
+
+    return resourceEvents.stream()
+      .flatMap(resourceEvent -> {
```
psmagin, line 163: @viacheslavkol, what do you think about filtering by non-instance resource type, to avoid the above processing for instance events?

viacheslavkol: I'm not sure what exactly do you meant.
If you are talking about ResourceEventType (ResourceEvent.type) - then data passed into this method has been already filtered (grouped by INDEX operation).
If you meant ResourceEvent.resourceName - then this method is used only for events that are all converted to instance events in Kafka listener so I think this is not necessary.

psmagin (approved):

Shans-Kaluhin (approved):

PR #313, opened 2022-11-15
MSEARCH-456 Call-number with same instance as anchor disappears

Shans-Kaluhin:
## Purpose
Call-number with same instance as anchor disappears

## Approach
Remove filtering intermediate items by `lower` and `upper` shelfKey. Because `lower` and `upper` depend on what anchor you are searching for;

**Example:**
Instance with Items ['A1', 'A2', 'A3']
If you are searching for 'A3' then the lower value will be 'A3'. And 'A1', 'A2' will be cut off.

This changes will not affect final size list, because we are sorting and trimming the results
 
### Learning
https://issues.folio.org/browse/MSEARCH-456

### DIFF
--- a/src/test/java/org/folio/search/service/browse/CallNumberBrowseResultConverterTest.java
+++ b/src/test/java/org/folio/search/service/browse/CallNumberBrowseResultConverterTest.java
@@ -161,7 +161,8 @@ class CallNumberBrowseResultConverterTest {
         forwardContext(), true, List.of(
           cnBrowseItem(instance("B1", "B2", "C2"), "B1"),
           cnBrowseItem(instance("B1", "B2", "C2"), "B2"),
-          cnBrowseItem(instance("C1", "C2"), "C1"))),
+          cnBrowseItem(instance("C1", "C2"), "C1"),
+          cnBrowseItem(2, "C2"))),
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/search/service/browse/CallNumberBrowseResultConverterTest.java
```diff
@@ -161,7 +161,8 @@ private static Stream<Arguments> testDataProvider() {
         forwardContext(), true, List.of(
           cnBrowseItem(instance("B1", "B2", "C2"), "B1"),
           cnBrowseItem(instance("B1", "B2", "C2"), "B2"),
-          cnBrowseItem(instance("C1", "C2"), "C1"))),
+          cnBrowseItem(instance("C1", "C2"), "C1"),
+          cnBrowseItem(2, "C2"))),
```
Shans-Kaluhin, line 165: 5 string above you can see that we have `instance("B1", "B2", "C2")` and `instance("C1", "C2")`
So, in the result we should have 2 instances with same call-number 'C2'. 
Before it was cut off, because `upper` value was 'C1'

psmagin (approved):

sherzodn (approved):

PR #314, opened 2022-11-17
MSEARCH-452 - MARC authority app - Create query that returns bib/instance records linked to authority record

sherzodn:
## Purpose
_MARC authority app - Create query that returns bib/instance records linked to authority record_

## Approach
When I click the number of titles link on the same row of one of the Authority results 
**Then** 
The query should only return instance/bib records that are linked to that authority record UUID (also known as has $9 populated with that UUID)
AND those results should display on the Inventory App > Instance results in list
Create a new Instance search option called Authority UUID SO that When the user goes to the results list

### Changes checklist
- [ ] Database schema changes
- [ ] Check logging.

### Learning
https://issues.folio.org/browse/MSEARCH-452


src/main/resources/model/instance.json
```diff
@@ -91,6 +91,11 @@
         "contributorTypeId": {
           "index": "keyword",
           "showInResponse": [ "search", "call-number-browse" ]
+        },
+        "authorityId": {
+          "index": "keyword",
+          "searchAliases": [ "authorityId" ],
+          "showInResponse": [ "search", "call-number-browse" ]
         }
```
psmagin, line 99: Please update the documentation with new search options

src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -64,6 +64,7 @@ void searchByInstances_parameterized_singleResult(String query, String value) th
     "uniformTitle all {value}, deja vu",
     "uniformTitle all {value}, déjà vu",
     "contributors.name all {value}, franks",
+    "contributors.authorityId == {value}, 11110000-fcf6-45cc-b6da-4420a61ef72c",
```
psmagin, line 67: Please add a test for the alias
`authorityId == {value}`

src/test/java/org/folio/search/controller/SearchInstanceIT.java
```diff
@@ -249,6 +250,8 @@ private static Stream<Arguments> testDataProvider() {
       arguments("contributors.name = {value}", "Anton*"),
       arguments("contributors.name = {value}", "*rmelen, Frank"),
 
+      arguments("contributors.authorityId == {value}", "55294032-fcf6-45cc-b6da-4420a61ef72c"),
```
psmagin, line 253: Please add a test for the alias
`authorityId == {value}`

psmagin (changes requested):
Please update the documentation with new search options

src/main/resources/model/instance.json
sherzodn: done

src/test/java/org/folio/search/controller/SearchInstanceIT.java
sherzodn: done

src/test/java/org/folio/search/controller/SearchInstanceIT.java
sherzodn: done

psmagin (approved):
Karate tests?

Shans-Kaluhin (approved):

PR #322, opened 2022-12-27
MSEARCH-479 Stabilize contributors browse integration tests

hasanepam:
## Purpose
Avoid random failure of Contributor browse integration test cases 

## Approach
Previously test method was sorting actual retrieved result's items field by only two fields: _Name_ and _ContributorNameTypeId_. Now three more field sorting added: _TotalRecords_, _AuthorityId_ and _Items.size()_

### TODOS and Open Questions
- [ ] Use GitHub checklists. When solved, check the box and explain the answer.

### Learning
[https://issues.folio.org/browse/MSEARCH-479](https://issues.folio.org/browse/MSEARCH-479)

### DIFF
--- a/src/test/java/org/folio/search/controller/BrowseContributorIT.java
+++ b/src/test/java/org/folio/search/controller/BrowseContributorIT.java
@@ -23,10 +23,10 @@ import static org.opensearch.index.query.QueryBuilders.matchAllQuery;
 import static org.opensearch.search.builder.SearchSourceBuilder.searchSource;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.ArrayList;
 import java.util.Map;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


src/test/java/org/folio/search/controller/BrowseContributorIT.java
```diff
@@ -300,8 +304,23 @@ void browseByContributor_parameterized(String query, String anchor, Integer limi
...
+    if (!CollectionUtils.isEmpty(expected.getItems())) {
+      assertEquals(expected.getItems().get(0), actual.getItems().get(0));
+      actual.getItems().remove(0);
+      List<InstanceContributorBrowseItem> actualItems = actual.getItems();
+      List<InstanceContributorBrowseItem> expectedItems = new ArrayList<>(expected.getItems());
+      expectedItems.remove(0);
+      assertTrue(expectedItems.containsAll(actualItems));
+    } else {
+      assertEquals(expected.getItems(), actual.getItems());
+    }
```
viacheslavkol, line 319: Why the first element is compared separately?
What's wrong with previous "assertThat(actual).isEqualTo(expected);" comparison?

hasanepam: - Because in search result first element should match exactly to the expected result.
- In previous "assertThat(actual).isEqualTo(expected);" comparison "items" order also should match as collection and when expected result's "items" differ by its order, tests failed. That's why first element is checked in order to check its order match as well as its value match, next elements' order can be skipped and thus _containsAll()_ used.

psmagin: But the order should match all collection. The only exception is items that have the same 'name', their order does not matter.

psmagin (approved):

sherzodn (approved):

PR #323, opened 2023-01-04
MSEARCH-468 Updated to jdk 17

hasanepam:
## Purpose
_The purpose of the PR is to increase version of JAVA to 17_


psmagin (approved):

src/main/java/org/folio/search/service/converter/preprocessor/AuthorityEventPreProcessor.java
```diff
@@ -95,8 +93,8 @@ private List<ResourceEvent> generateResourceEvents(ResourceEvent event, Resource
   }
 
   private List<ResourceEvent> createResourceEvents(ResourceEvent event, String type, String name,
-                                                   AtomicInteger counter, ResourceEventType eventType,
-                                                   Map<String, Object> body) {
+    AtomicInteger counter, ResourceEventType eventType,
```
sherzodn, line 96: should we keep tabs here?

sherzodn (approved):

PR #325, opened 2023-01-10
MSEARCH-487 Added folio-service-tools-spring libraries

hasanepam:
## Purpose
_Use library classes instead of duplicated project ones._

## Approach
_Removed unneeded actual and test classes, changed properties to match library environment variable uses_


src/main/java/org/folio/search/service/KafkaConstants.java
```diff
@@ -0,0 +1,7 @@
+package org.folio.search.service;
+
+public class KafkaConstants {
+  public static final String EVENT_LISTENER_ID = "mod-search-events-listener";
```
Shans-Kaluhin, line 4: We can add private constructor just to remove code smell. 
And maybe we should move constants to utils package. Out of services?
```suggestion
  private KafkaConstants() {}
  public static final String EVENT_LISTENER_ID = "mod-search-events-listener";
```

hasanepam: Fixed as suggested.

Shans-Kaluhin (approved):

PR #330, opened 2023-02-16
MSEARCH-299 Logging improvement

shorasul96:
## Purpose
_Implementing and improving logs for more productive troubleshooting_

## Approach
_Logging is implemented in business logic steps, most edge scenarios, conditional business logics and exceptional cases._

### Changes checklist
- [ ] Database schema changes
- [x] Check logging.

### DIFF
--- a/src/main/java/org/folio/search/service/converter/preprocessor/AuthorityEventPreProcessor.java
+++ b/src/main/java/org/folio/search/service/converter/preprocessor/AuthorityEventPreProcessor.java
@@ -64,18 +64,17 @@ public class AuthorityEventPreProcessor implements EventPreProcessor {
    */
   @Override
   public List<ResourceEvent> process(ResourceEvent event) {
-    log.debug("process:: by [event: {}]", event);
+    log.debug("process:: by [id: {}, tenant: {}, resourceType: {}]",
+      event.getId(), event.getTenant(), event.getType());
 
     if (event.getType() == ResourceEventType.UPDATE) {
-      log.debug("process:: eventType == UPDATE");
       return getResourceEventsToUpdate(event);
     }
-    log.debug("process:: eventType != UPDATE");
     return getResourceEvents(event, event.getType());
   }
 
   private List<ResourceEvent> getResourceEvents(ResourceEvent event, ResourceEventType eventType) {
-    log.debug("getResourceEvents:: by [event: {}, type: {}]", event, eventType);
+    log.debug("getResourceEvents:: by [id: {}, tenant: {}, type: {}]", event.getId(), event.getTenant(), eventType);
 
     var isCreateOperation = isCreateOperation(eventType);
     var events = generateResourceEvents(event, eventType, isCreateOperation ? getNewAsMap(event) : getOldAsMap(event));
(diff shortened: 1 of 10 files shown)


hasanepam (approved):

src/main/java/org/folio/search/service/converter/preprocessor/AuthorityEventPreProcessor.java
```diff
@@ -61,15 +64,25 @@ public void init() {
...
-           ? getResourceEventsToUpdate(event)
-           : getResourceEvents(event, event.getType());
+    log.debug("process:: by [event: {}]", event);
+
+    if (event.getType() == ResourceEventType.UPDATE) {
+      log.debug("process:: eventType == UPDATE");
+      return getResourceEventsToUpdate(event);
+    }
+    log.debug("process:: eventType != UPDATE");
+    return getResourceEvents(event, event.getType());
```
viacheslavkol, line 74: ResourceEvent is too big to put it in logs. Only basic info could be logged (Same for other places where big object is loggged).
You shouldn't write multiple methods for different event types, you could just put eventType into a log

src/main/java/org/folio/search/service/metadata/LocalResourceProvider.java
```diff
@@ -51,13 +52,13 @@ public Map<String, SearchFieldType> getSearchFieldTypes() {
...
   private void loadResourceDescriptions() {
     try {
+      log.debug("loadResourceDescriptions:: Attempting to load descriptions");
       this.resourceDescriptions = Stream.of(patternResolver.getResources(RESOURCE_DESCRIPTIONS_LOCATION_PATTERN))
         .filter(Resource::isReadable)
         .map(this::loadResourceDescription)
         .filter(Objects::nonNull)
         .toList();
     } catch (IOException e) {
-      log.error("Failed to read models [pattern: {}]", RESOURCE_DESCRIPTIONS_LOCATION_PATTERN);
```
viacheslavkol, line 60: Why are this and next error logs removed?

src/main/java/org/folio/search/utils/LogUtils.java
```diff
@@ -0,0 +1,43 @@
+package org.folio.search.utils;
+
+import java.util.Collection;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class LogUtils {
```
viacheslavkol, line 8: I'd say it's not necessary. We usually log only size of list. And even list with size = 2 may be too big (if large objects stored) to put in log.
It should be decided individually when logging.

src/main/java/org/folio/search/service/IndexService.java
```diff
@@ -47,15 +48,18 @@ public class IndexService {
...
       "Index cannot be created for the resource because resource description is not found.");
 
     var index = getIndexName(resourceName, tenantId);
     var settings = settingsHelper.getSettings(resourceName);
     var mappings = mappingHelper.getMappings(resourceName);
 
-    log.info("Creating index for resource [resource: {}, tenant: {}, mappings: {}, settings: {}]",
-      resourceName, tenantId, mappings, settings);
+    log.info("Attempts to create index by [indexName: {}, mappings: {}, settings: {}]",
+      index, mappings, settings);
```
viacheslavkol, line 62: We have multitenancy in system so It's needed to add tenantId, where it's available

viacheslavkol (commented):
I believe we should add 'info' level logs for APIs' inputs/outputs to have some basic info in there.
Also I'd suggest returning removed warn level logs (or why were they removed?)

viacheslavkol (approved):

PR #334, opened 2023-03-17
MSEARCH-503 Call number browse: do an additional query for preceding records when none returned

Shans-Kaluhin:
## Approach
- Do an additional query for preceding records when none returned

### Learning
https://issues.folio.org/browse/MSEARCH-503


ArtHov (approved):

psmagin (approved):

src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
```diff
@@ -76,6 +82,27 @@ protected String getValueForBrowsing(CallNumberBrowseItem browseItem) {
...
 
+  private BrowseResult<CallNumberBrowseItem> additionalPrecedingRequests(BrowseRequest request,
+                                                                         BrowseContext context,
+                                                                         SearchSourceBuilder precedingQuery) {
+    BrowseResult<CallNumberBrowseItem> precedingResult = BrowseResult.empty();
+
+    while (precedingResult.getRecords().isEmpty()) {
+      int offset = precedingQuery.from() + precedingQuery.size();
+      log.debug("additionalPrecedingRequests:: request offset {}", offset);
+      precedingQuery.from(offset);
```
viacheslavkol, line 93: Shouldn't we change `to` parameter instead for preceding request?
Also there's a case when numeric representation for a call number may be the same for over 300 records and query size has a minimum of 25 so maybe it'd good to have it increased on "additionalPrecedingRequests" so we don't make a call 12 times?

Shans-Kaluhin: - Query doesn't have `to` value. Only `from` and `size`
- Added BROWSE_CN_ADDITIONAL_REQUEST_SIZE env variable

hasanepam (approved):

PR #338, opened 2023-04-17
    MSEARCH-513  An error occurs when searching by the first call num…

pawelmol:
## Approach
Added limit on while loop in CallNumberBrowseService.additionalPrecedingRequests(). When there is no precedding call numbers (very first call number) this loop run to infinity (to exception).

### Learning
[_Describe the research stage. Add links to blog posts, patterns, libraries or addons used to solve this problem._](https://issues.folio.org/browse/MSEARCH-513)

### DIFF
--- a/src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
+++ b/src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
@@ -25,7 +25,7 @@ import org.springframework.stereotype.Service;
 public class CallNumberBrowseService extends AbstractBrowseService<CallNumberBrowseItem> {
 
   private static final int ADDITIONAL_REQUEST_SIZE = 100;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
```diff
@@ -25,6 +25,7 @@
 public class CallNumberBrowseService extends AbstractBrowseService<CallNumberBrowseItem> {
 
   private static final int ADDITIONAL_REQUEST_SIZE = 100;
+  private static final int ADDITIONAL_REQEST_SIZE_MAX = 500;
```
psmagin, line 28:
```suggestion
  private static final int ADDITIONAL_REQUEST_SIZE_MAX = 500;
```
Please fix the typo

ArtHov (approved):

PR #339, opened 2023-04-29
MSEARCH-514 Add new field to index to support CN browse by types

psmagin:
## Purpose
Add a new field to the instance index to support call-numbers browse by types

## Approach
- Add typedCallNumber to index mapping
- Extend item schema to have callNumberTypeId
- Add processor to the new field in index, use reference service to load `local` call-number types, and implement new call-number to long conversion method to have reserved 1st digit for CN type
- Add new permission for system-user

### Changes checklist
- [ ] Database schema changes
- [ ] Check logging


Shans-Kaluhin (approved):

hasanepam (approved):

src/main/java/org/folio/search/utils/CallNumberUtils.java
```diff
@@ -0,0 +1,182 @@
...
+    long startVal = convertChar(firstPosition, CN_MAX_CHARS);
+    return callNumberToLong(callNumber, startVal, CN_MAX_CHARS - 1);
+  }
+
+  private static long callNumberToLong(String callNumber, long startVal, int maxChars) {
+    var cleanCallNumber = cleanCallNumber(callNumber, maxChars);
+    if (StringUtils.isBlank(cleanCallNumber)) {
+      return 0L;
+    }
+    long result = startVal;
```
viacheslavkol, line 149: image

src/main/java/org/folio/search/utils/CallNumberUtils.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.stream.Stream;
+import lombok.experimental.UtilityClass;
+import org.apache.commons.lang3.RegExUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.util.Assert;
+
+@UtilityClass
+public class CallNumberUtils {
+
+  private static final int CN_MAX_CHARS = 10;
```
viacheslavkol, line 21: image

viacheslavkol (changes requested):

src/main/java/org/folio/search/utils/CallNumberUtils.java
psmagin: Check line 140 in the file, it does what described

src/main/java/org/folio/search/utils/CallNumberUtils.java
psmagin: In method on line 139 there is 'CN_MAX_CHARS-1'

viacheslavkol (approved):

PR #346, opened 2023-05-10
MSEARCH-518  Adjust GET /search/authorities to not include titles numberOfTitles

ArtHov:
## Purpose
To increase the performance of automated linking we need to make GET /search/authorities more performant.

## Approach
Add a new boolean query parameter includeNumberOfTitles to indicate whether the number of titles should be counted.

### Changes checklist
- [ ] Database schema changes
- [ ] Check logging.

### TODOS and Open Questions
- [ ] Use GitHub checklists. When solved, check the box and explain the answer.

### Learning
_Describe the research stage. Add links to blog posts, patterns, libraries or addons used to solve this problem._

### DIFF
--- a/src/test/java/org/folio/search/service/converter/ElasticsearchDocumentConverterTest.java
+++ b/src/test/java/org/folio/search/service/converter/ElasticsearchDocumentConverterTest.java
@@ -10,12 +10,8 @@ import static org.folio.search.utils.TestUtils.array;
 import static org.folio.search.utils.TestUtils.mapOf;
 import static org.folio.search.utils.TestUtils.searchResult;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyList;
 import static org.mockito.ArgumentMatchers.anyMap;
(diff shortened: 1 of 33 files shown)


psmagin (commented):
Hello @ArtHov, I have some thoughts about potential changes that could possibly be omitted. Have you considered moving the processing done by SearchResponsePostProcessor from the converter to `SearchService` and `AbstractBrowseService`? I believe this could be a good approach since post-processing is outside the scope of conversion. What do you think?

ArtHov:
I also have the same opinion as the processing of additional logic in the converter is a little confusing. I will make a refactoring of that part. Thanks.

viacheslavkol (approved):

src/test/java/org/folio/search/service/converter/ElasticsearchDocumentConverterTest.java
```diff
@@ -63,6 +63,8 @@ class ElasticsearchDocumentConverterTest {
   private SearchHits searchHits;
   @Mock
   private SearchHit searchHit;
+  @Mock
+  private SearchResponsePostProcessor<?> searchResponsePostProcessor;
```
psmagin, line 67: As I can see the mock is never used as well as map:
image

PR #358, opened 2023-05-25
MSEARCH-523 Include call number type id from Holdings in Items record when not specified

viacheslavkol:
## Purpose
For browsing by call number type, the Item needs to appear under the Call number type specified either in the Item record or inherited from the Holdings record 
[MSEARCH-523](https://issues.folio.org/browse/MSEARCH-523)

## Approach
- Use item.effectiveCallNumberComponents.typeId instead of item.itemLevelCallNumberTypeId for item call number type
- Add item.effectiveCallNumberComponents.typeId to search/browse responses

### Changes checklist
- [x] API paths, methods, request or response bodies changed, added, or removed
- [ ] Database schema changes
- [x] Interface versions changes
- [ ] Check logging.

### DIFF
--- a/src/test/java/org/folio/search/service/setter/item/ItemTypedCallNumberProcessorTest.java
+++ b/src/test/java/org/folio/search/service/setter/item/ItemTypedCallNumberProcessorTest.java
@@ -1,6 +1,5 @@
 package org.folio.search.service.setter.item;
 
-import static org.apache.commons.collections4.SetUtils.emptySet;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.folio.search.client.InventoryReferenceDataClient.ReferenceDataType.CALL_NUMBER_TYPES;
 import static org.folio.search.model.client.CqlQueryParam.SOURCE;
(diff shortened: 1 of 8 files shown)


ArtHov (approved):

src/test/java/org/folio/search/service/setter/item/ItemTypedCallNumberProcessorTest.java
```diff
@@ -66,8 +66,6 @@ void getFieldValue_emptyCallNumberAfterProcessing(String effectiveShelvingOrder,
 
   @Test
   void getFieldValue_emptyCallNumberIfNotSystemOrNotLocal() {
-    when(referenceDataService.fetchReferenceData(CALL_NUMBER_TYPES, SOURCE, LOCAL_CALL_NUMBER_TYPES_SOURCES))
-      .thenReturn(emptySet());
```
psmagin, line 70: Why the mock is not needed anymore?

viacheslavkol: I'm not sure was it was needed at all, Mockito returns empty collection for collections by default.
Also test was failing with this stubbing for me (saying that it's unnecessary).

psmagin: Strange. Okay

psmagin (approved):

PR #384, opened 2023-07-27
feat(consortium): implement indexing for consortium (kafka part)

psmagin:
## Purpose
Implement indexing for consortium (kafka part)

## Approach
* add new topic
* add logic to publish instance ids related to consortium to the topic
* add logic to consume these ids and index to search engine

### Changes checklist
- [ ] Database schema changes
- [ ] Check logging.


src/main/java/org/folio/search/service/consortium/ConsortiumInstanceService.java
```diff
@@ -164,4 +190,17 @@ private boolean isCentralTenant(String tenantId) {
     var centralTenant = consortiumTenantService.getCentralTenant(tenantId);
     return centralTenant.isPresent() && centralTenant.get().equals(tenantId);
   }
+
+  private String getCentralId() {
+    return consortiumTenantService.getCentralTenant(context.getTenantId()).orElse(null);
```
viacheslavkol, line 195: Shouldn't we just use 'TenantProvider' here, which returns initially passed tenantId in case central is absent?
Or is there some reason for passing 'null'?

src/test/java/org/folio/search/controller/SearchInstanceFilterIT.java
```diff
@@ -328,7 +328,7 @@ private static Stream<Arguments> facetQueriesProvider() {
       "tenantId", "shared");
     return Stream.of(
       arguments("id=*", allFacets, mapOf(
-        "shared", facet(facetItem("false", 3), facetItem("true", 2)),
+        "shared", facet(facetItem("false", 5)),
```
viacheslavkol, line 331: Why are there no "true" values?
Test data should be changed in a scope of MSEARCH-562 in order to have them?

psmagin: Yes, we should implement some setup for consortia tests, where we will have more than 1 tenant enabled.

viacheslavkol: We currently have 2 tenants setup for integration tests, see BaseIntegrationTest:enableTenant usage.
But possibly we lack 'shared' logic

src/main/java/org/folio/search/service/ResourceService.java
```diff
@@ -129,20 +129,26 @@ public FolioIndexOperationResponse indexConsortiumInstances(List<ConsortiumInsta
...
-        allResourceEvents.putAll(multiTenantSearchDocumentConverter.convert(resourceEvents));
-        return null;
-      });
+    if (log.isDebugEnabled()) {
+      var invalidInstances = ListUtils.subtract(consortiumInstances, validConsortiumInstances);
+      log.debug("Skip indexing consortium instances [{}]", invalidInstances);
     }
 
-    return indexSearchDocuments(allResourceEvents);
+    var centralTenant = consortiumTenantService.getCentralTenant(validConsortiumInstances.get(0).getInstanceId())
```
viacheslavkol, line 142: Why is instanceId passed here?
Are you sure /user-tenants will return central tenant in such case?

viacheslavkol (approved):

PR #393, opened 2023-08-08
MSEARCH-559: Call Numbers Browse with call number type shows irrelevant result corrected

hasanepam:
## Purpose
Fix irrelevant result on browsing call numbers by call number type

## Approach
Add filter for search result to eliminate irrelevant search results which came together with instance.

### Learning
(MSEARCH-557)[https://issues.folio.org/browse/MSEARCH-557]

### DIFF
--- a/src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
+++ b/src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
@@ -4,6 +4,7 @@ import static java.lang.Boolean.TRUE;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static org.apache.commons.collections.CollectionUtils.isEmpty;
+import static org.folio.search.utils.CallNumberUtils.excludeIrrelevantResultItems;
 import static org.folio.search.utils.CollectionUtils.mergeSafelyToList;
 
 import java.util.ArrayList;
@@ -41,6 +42,9 @@ public class CallNumberBrowseService extends AbstractBrowseService<CallNumberBro
     var searchResponse = searchRepository.search(request, searchSource);
     var browseResult = callNumberBrowseResultConverter.convert(searchResponse, context, isBrowsingForward);
     var records = browseResult.getRecords();
+    if (request.getRefinedCondition() != null) {
+      browseResult.setRecords(excludeIrrelevantResultItems(request.getRefinedCondition(), browseResult.getRecords()));
+    }
     return new BrowseResult<CallNumberBrowseItem>()
       .records(trim(records, context, isBrowsingForward))
       .totalRecords(browseResult.getTotalRecords())
@@ -80,6 +84,12 @@ public class CallNumberBrowseService extends AbstractBrowseService<CallNumberBro
       highlightMatchingCallNumber(context, callNumber, succeedingResult);
     }
 
+    if (request.getRefinedCondition() != null) {
+      precedingResult.setRecords(excludeIrrelevantResultItems(request.getRefinedCondition(),
+        precedingResult.getRecords()));
+      succeedingResult.setRecords(excludeIrrelevantResultItems(request.getRefinedCondition(),
+        succeedingResult.getRecords()));
+    }
     return new BrowseResult<CallNumberBrowseItem>()
       .totalRecords(precedingResult.getTotalRecords() + succeedingResult.getTotalRecords())
       .prev(getPrevBrowsingValue(precedingResult.getRecords(), context, false))
--- a/src/main/java/org/folio/search/utils/CallNumberUtils.java
+++ b/src/main/java/org/folio/search/utils/CallNumberUtils.java
@@ -3,6 +3,7 @@ package org.folio.search.utils;
 import static java.util.Collections.unmodifiableMap;
 import static java.util.Locale.ROOT;
 import static java.util.stream.Collectors.joining;
+import static org.folio.search.utils.CollectionUtils.distinctByKey;
 
 import java.util.HashMap;
 import java.util.List;
@@ -13,6 +14,8 @@ import java.util.stream.Stream;
 import lombok.experimental.UtilityClass;
 import org.apache.commons.lang3.RegExUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.folio.search.domain.dto.CallNumberBrowseItem;
+import org.folio.search.model.types.CallNumberType;
 import org.springframework.util.Assert;
 
 @UtilityClass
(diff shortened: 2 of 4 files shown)


src/main/java/org/folio/search/utils/CallNumberUtils.java
```diff
@@ -144,6 +147,25 @@ public static Long getCallNumberAsLong(String callNumber, int firstPosition) {
...
+    records = records
+      .stream()
+      .filter(distinctByKey(r -> r.getInstance().getId()))
+      .toList();
+    records
+      .forEach(r -> r.getInstance().setItems(r.getInstance().getItems()
+        .stream()
+        .filter(i -> CallNumberType.fromId(i.getEffectiveCallNumberComponents().getTypeId())
+          .equals(CallNumberType.fromName(refinedCondition)))
+        .toList()));
```
viacheslavkol, line 165: Looks like we can do it in one stream sequence.
And why do wee need to filter records by instanceId at all?

src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
```diff
@@ -41,6 +42,7 @@ protected BrowseResult<CallNumberBrowseItem> browseInOneDirection(BrowseRequest
     var searchResponse = searchRepository.search(request, searchSource);
     var browseResult = callNumberBrowseResultConverter.convert(searchResponse, context, isBrowsingForward);
     var records = browseResult.getRecords();
+    browseResult.setRecords(excludeIrrelevantResultItems(request.getRefinedCondition(), browseResult.getRecords()));
```
viacheslavkol, line 45:
```suggestion
    browseResult.setRecords(excludeIrrelevantResultItems(request.getRefinedCondition(), records));
```

psmagin (changes requested):
Please write a test where we can check pagination: returned by the elastic number of records is higher than limit in the API request

src/main/java/org/folio/search/utils/CallNumberUtils.java
```diff
@@ -144,6 +150,60 @@ public static Long getCallNumberAsLong(String callNumber, int firstPosition) {
...
+        .stream()
+        .filter(h -> isInGivenType(refinedCondition, h.getCallNumber()))
+        .toList()));
+    var result = new ArrayList<>(records
+      .stream()
+      .filter(r -> r.getInstance().getItems()
+        .stream()
+        .anyMatch(i -> r.getFullCallNumber() == null
+          || i.getEffectiveCallNumberComponents().getCallNumber().equals(r.getFullCallNumber())))
+      .toList());
```
viacheslavkol, line 176: What is the purpose of this filtration?
And the next one for holdings

src/test/java/org/folio/search/controller/BrowseCallNumberIrrelevantResultTest.java
```diff
@@ -0,0 +1,196 @@
...
+      cnBrowseItem(instance("instance #01"), "308 H977", true)
+    )).next("3308 H977");
+    assertThat(actual).isEqualTo(expected);
+    assertThat(actual.getItems()).hasSizeLessThanOrEqualTo(limit);
+  }
+
+  private static Instance[] instances() {
+    return new Instance[] {
+      instance(callNumberBrowseInstanceData()),
+      instanceNew(additionalCallNumberBrowseInstanceData()),
```
psmagin, line 92: Why it is `new` instance? What do you mean with that?

src/test/java/org/folio/search/controller/BrowseCallNumberIrrelevantResultTest.java
```diff
@@ -0,0 +1,196 @@
...
+      .param("highlightMatch", "true")
+      .param("precedingRecordsCount", "1")
+      .param("expandAll", "true");
+    var actual = parseResponse(doGet(request), CallNumberBrowseResult.class);
+    var expected = cnBrowseResult(3, List.of(
+      cnBrowseItem(instance("instance #01"), "308 H977", true)
+    )).next("3308 H977");
+    assertThat(actual).isEqualTo(expected);
+    assertThat(actual.getItems()).hasSizeLessThanOrEqualTo(limit);
+  }
```
psmagin, line 87: There is no sense in the test as it works with `callNumber` and not `typedCallNumber` .
You need to create a test where based on query elastic search could return more than a limit. With the test we should understand that pagination works fine and we get full page even with you filtration.

psmagin (changes requested):

src/main/java/org/folio/search/utils/CallNumberUtils.java
hasanepam: when one instance has two or more call numbers in different types, browsing one of those call numbers by type will bring all of the instance call numbers, because call number browse item includes instance information for each matching result, and instance will bring all of its call numbers.
All instance call numbers will be included in the browse result because of one of them .

src/main/java/org/folio/search/utils/CallNumberUtils.java
```diff
@@ -144,6 +153,75 @@ public static Long getCallNumberAsLong(String callNumber, int firstPosition) {
...
+      r.getInstance().setHoldings(getHoldingsFiltered(callNumberType, r));
+    });
+    return records;
+  }
+
+  @NotNull
+  private static List<@Valid Holding> getHoldingsFiltered(String callNumberType, CallNumberBrowseItem item) {
+    return item.getInstance().getHoldings()
+      .stream()
+      .filter(h -> isInGivenType(callNumberType, h.getCallNumber()))
```
psmagin, line 192: Why don't you check it by callNumberTypeId like for items?

psmagin: Why do we need to filter both holdings and items? Do we get call numbers from both holdings and items while converting elastic search responses?

src/test/java/org/folio/search/controller/BrowseTypedCallNumberIrrelevantResultTest.java
```diff
@@ -0,0 +1,173 @@
...
+import org.folio.search.domain.dto.ItemEffectiveCallNumberComponents;
+import org.folio.search.support.base.BaseIntegrationTest;
+import org.folio.search.utils.CallNumberUtils;
+import org.folio.spring.test.type.IntegrationTest;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+@IntegrationTest
+class BrowseTypedCallNumberIrrelevantResultTest extends BaseIntegrationTest {
```
psmagin, line 32: If it is an integration test then rename it to have IT prefix.

src/test/java/org/folio/search/controller/BrowseTypedCallNumberIrrelevantResultTest.java
```diff
@@ -0,0 +1,173 @@
...
+      cnBrowseItem(INSTANCE_MAP.get("instance #11"), "308 H972"),
+      cnBrowseItem(INSTANCE_MAP.get("instance #11"), "308 H973"),
+      cnBrowseItem(INSTANCE_MAP.get("instance #11"), "308 H974"),
+      cnBrowseItem(INSTANCE_MAP.get("instance #09"), "308 H977", true),
+      cnBrowseItem(INSTANCE_MAP.get("instance #01"), "308 H978"),
+      cnBrowseItem(INSTANCE_MAP.get("instance #10"), "308 H979"),
+      cnBrowseItem(INSTANCE_MAP.get("instance #02"), "308 H980"),
+      cnBrowseItem(INSTANCE_MAP.get("instance #08"), "308 H981")
+    ));
+    expected.setItems(CallNumberUtils.excludeIrrelevantResultItems("dewey", expected.getItems()));
```
psmagin, line 71: Why we can't just have `expected` in final form and do `expected.setItems(CallNumberUtils.excludeIrrelevantResultItems("dewey", expected.getItems()))` ?

src/main/java/org/folio/search/utils/CallNumberUtils.java
hasanepam: 1. Because in holdings there is no type information(id or name of type).
2. I am filtering both items and holdings call numbers because in the ticket it is being claimed that irrelevant call numbers from holdings are coming up while browsing.

psmagin (approved):

mukhiddin-yusuf (approved):

PR #416, opened 2023-09-12
MSEARCH-557: Call number browse with call number type shows irrelevant result fix

hasanepam:
## Purpose
Fix related to browse in one direction

### Learning
[MSEARCH-557](https://issues.folio.org/browse/MSEARCH-557)


andrei-razumnov (approved):

psmagin (changes requested):
Please rename PR with meaningful name. 
Please add test to the case you are fixed, because currently there is no understanding why this fix done.

src/main/java/org/folio/search/utils/CallNumberUtils.java
```diff
@@ -178,19 +174,13 @@ public static List<CallNumberBrowseItem> excludeIrrelevantResultItems(String cal
...
-    return item.getInstance().getHoldings()
+    records.forEach(r -> r.getInstance().setItems(getItemsFiltered(callNumberType, r)));
+    return records
       .stream()
-      .filter(h -> isInGivenType(callNumberType, h.getCallNumber()))
-      .toList();
+      .filter(r -> r.getInstance().getItems()
+        .stream()
+        .anyMatch(i -> r.getFullCallNumber() == null
+          || i.getEffectiveCallNumberComponents().getCallNumber().equals(r.getFullCallNumber()))).toList();
```
hasanepam, line 183: This check is added in order to exclude non-matching call numbers, which an resulting instance, with matching call number, may include.

psmagin (approved):

PR #421, opened 2023-09-21
MSEARCH-557

hasanepam:
## Purpose
Avoid NPE which is being thrown while browsing call numbers (all) using one letter (s)

## Approach
Checked for null before further processing

### Learning
[MSEARCH-557](https://issues.folio.org/browse/MSEARCH-557)


viacheslavkol:
Is implemented in a scope of [mod-search/pull/424](https://github.com/folio-org/mod-search/pull/424)

PR #433, opened 2023-10-03
fix(cn-browse): Fix call number not found for multiple titles

viacheslavkol:
## Purpose
Fix call number not found for multiple titles
[MSEARCH-598](https://issues.folio.org/browse/MSEARCH-598)

## Approach
- Do not filter out results with no instance record

### Changes checklist
- [ ] Database schema changes
- [ ] Check logging.


psmagin (approved):

src/main/java/org/folio/search/utils/CallNumberUtils.java
andrei-razumnov, line 1: Can use `CollectionUtils.isEmpty(records)`

andrei-razumnov, line 1: Can use `CollectionUtils.isEmpty(records)` here

andrei-razumnov (approved):

PR #440, opened 2023-10-11
Msearch 589 "Exact full match" operator doesn't work for alias fields.

TsaghikKhachatryan:
## Purpose
The "Exact full match" operator doesn't work for alias fields.

## Approach
Alias fields are replaced with a 'plan_' prefix.

Query **before** changes:

    "multi_match" : {
      "query" : "Sit how you want / Robin Richardson.",
      "fields" : [
      "alternativeTitles.alternativeTitle.*^1.0",
      "contributors.name^1.0",
      "identifiers.value^1.0",
      "indexTitle.*^1.0",
      "series.value.*^1.0",
      "title.*^1.0"
    ],
    "type": "phrase"
    }
  

Query **after** changes:

   "multi_match" : {
      "query" : "Sit how you want / Robin Richardson.",
      "fields" : [
      "alternativeTitles.**plain_alternativeTitle**^1.0",
      "contributors.**plain_name**^1.0",
      "identifiers.**plain_value**^1.0",
      "**plain_indexTitle**^1.0",
      "**plain_title**^1.0",
      "series.**plain_value**^1.0"
    ],
    "type": "phrase"
    }

### Learning
https://issues.folio.org/browse/MSEARCH-589

### DIFF
--- a/src/main/java/org/folio/search/cql/builders/ExactTermQueryBuilder.java
+++ b/src/main/java/org/folio/search/cql/builders/ExactTermQueryBuilder.java
@@ -8,7 +8,6 @@ import static org.opensearch.index.query.QueryBuilders.multiMatchQuery;
 import static org.opensearch.index.query.QueryBuilders.scriptQuery;
 import static org.opensearch.index.query.QueryBuilders.termQuery;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import org.folio.search.utils.SearchUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/search/cql/builders/ExactTermQueryBuilder.java
```diff
@@ -25,7 +26,10 @@ public class ExactTermQueryBuilder extends FulltextQueryBuilder {
 
   @Override
   public QueryBuilder getQuery(Object term, String resource, String... fields) {
-    return multiMatchQuery(term, fields).type(PHRASE);
+    List<String> updatedFields = Arrays.stream(fields)
+      .map(field -> updatePathForTermQueries(resource, field))
```
psmagin, line 30: This adjustment have to be done only when we have `/string` modifier. `==` operator must work as before. Changes should be only with `==\string` operator with modifier. 
The same like we have there:
image

psmagin (changes requested):

psmagin:
Also please follow conventional commit format as described on a page: https://wiki.folio.org/display/FOLIJET/Conventional+Commit+Initiative

PR #441, opened 2023-10-12
MSEARCH-589 "Exact full match" operator doesn't work for alias fields

TsaghikKhachatryan:
## Purpose
The "Exact full match" operator ( ==/string )doesn't work for alias fields.

## Approach
Alias fields are replaced with a 'plan_' prefix.

### Changes checklist

**Query before changes:**

  "query" : "Sit how you want / Robin Richardson.",
  "fields" : [
  "alternativeTitles.alternativeTitle.*^1.0",
  "contributors.name^1.0",
  "identifiers.value^1.0",
  "indexTitle.*^1.0",
  "series.value.*^1.0",
  "title.*^1.0"
],
"type": "phrase"
}

**Query after changes:**

"multi_match" : {
"query" : "Sit how you want / Robin Richardson.",
"fields" : [
"alternativeTitles.plain_alternativeTitle^1.0",
"contributors.plain_name^1.0",
"identifiers.plain_value^1.0",
"plain_indexTitle^1.0",
"plain_title^1.0",
"series.plain_value^1.0"
],
"type": "phrase"
}

### Learning
https://issues.folio.org/browse/MSEARCH-589

### DIFF
--- a/src/main/java/org/folio/search/cql/CqlTermQueryConverter.java
+++ b/src/main/java/org/folio/search/cql/CqlTermQueryConverter.java
@@ -21,6 +21,7 @@ import org.folio.search.exception.ValidationException;
 import org.folio.search.model.metadata.PlainFieldDescription;
 import org.folio.search.service.metadata.LocalSearchFieldProvider;
 import org.folio.search.service.metadata.SearchFieldProvider;
+import org.folio.search.utils.SearchUtils;
 import org.opensearch.index.query.QueryBuilder;
 import org.springframework.stereotype.Component;
 import org.springframework.util.Assert;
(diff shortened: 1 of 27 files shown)


src/main/java/org/folio/search/cql/CqlTermQueryConverter.java
```diff
@@ -89,7 +91,16 @@ public QueryBuilder getQuery(CQLTermNode termNode, String resource) {
...
+    var modifiers = termNode.getRelation().getModifiers().stream()
+      .map(Modifier::getType)
+      .toList();
+
     if (CollectionUtils.isNotEmpty(fieldsList)) {
+      if (modifiers.contains(STRING_MODIFIER)) {
+        var updatedFieldsList = getUpdatedFields(fieldsList);
+        return termQueryBuilder.getQuery(searchTerm, resource, updatedFieldsList.toArray(String[]::new));
+      }
+
```
psmagin, line 103: In such implementation, you modify all possible operators, but this `string` modifier could work and impact on query only for the "==" operator. 
I expect that `TermQueryBuilder.getQuery` method will be modified to have a new parameter with a list of modifiers. It will help to separate logic of query creation for each operator and will open opportunities for future implementation for other modifiers.

psmagin (changes requested):

psmagin (approved):

viacheslavkol (approved):

PR #455, opened 2023-10-25
fix(cn-browse): Fix result filtering for multiple items.effectiveLocationId

viacheslavkol:
## Purpose
Fix irrelevant results on multiple location filters

## Approach
Fix result filtering for multiple items.effectiveLocationId

### Changes checklist
- [ ] Database schema changes
- [ ] Check logging.

### DIFF
--- a/src/main/java/org/folio/search/service/consortium/ConsortiumSearchHelper.java
+++ b/src/main/java/org/folio/search/service/consortium/ConsortiumSearchHelper.java
@@ -8,12 +8,14 @@ import static org.opensearch.index.query.QueryBuilders.boolQuery;
 import static org.opensearch.index.query.QueryBuilders.termQuery;
 
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/search/service/consortium/ConsortiumSearchHelper.java
```diff
@@ -187,14 +189,35 @@ private Optional<TermQueryBuilder> getBrowseTenantFilter(BrowseContext context)
...
+      : null;
+  }
+
+  private static Stream<TermQueryBuilder> getTermFiltersForKey(QueryBuilder filter, String filterKey) {
+    if (filter instanceof TermQueryBuilder termFilter && termFilter.fieldName().equals(filterKey)) {
+      return Stream.of(termFilter);
+    } else if (filter instanceof BoolQueryBuilder boolFilter) {
+      return boolFilter.should().stream().map(shouldFilter -> getTermFilterForKey(shouldFilter, filterKey));
+    }
+    return null;
```
mukhiddin-yusuf, line 217: ```Stream.empty()``` could be returned as I think ```.flatMap()``` in any way would convert to empty stream if mapper returns ```null``` (according to javadoc), or I think there is no need to ```.filter(Objects::nonNull)```

Also, ```getTermFilterForKey(...)``` could also be changed to return Stream

mukhiddin-yusuf (approved):

psmagin (approved):

TsaghikKhachatryan (approved):

PR #460, opened 2023-11-06
MSEARCH-617 Ignore hard-delete domain authority events

GeloPakDev1:
## Purpose
Ignore hard-delete domain authority events

## Approach
dd filter to Kafka listener to skip hard-delete subtype for authority domain events

### Learning
Reference to AuthorityChangeFilterStrategy class in mod-entities-links


src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+      case UPDATE -> {
+        if (resourceHasChanges(resourceEvent)) {
+          yield false;
+        } else {
+          log.info("Skip message. No need to process update event for authority resource");
+          yield true;
+        }
+      }
+      case CREATE, REINDEX -> true;
+      case DELETE, DELETE_ALL -> {
```
mukhiddin-yusuf, line 38: Not sure about the ```DELETE_ALL```, but I think both authority soft and hard delete operations would have event type ```DELETE```, the differentiation would be based on the subtype which would be encoded in a header of the event.

src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+    return switch (resourceEvent.getType()) {
+      case UPDATE -> {
+        if (resourceHasChanges(resourceEvent)) {
+          yield false;
+        } else {
+          log.info("Skip message. No need to process update event for authority resource");
+          yield true;
+        }
+      }
+      case CREATE, REINDEX -> true;
```
mukhiddin-yusuf, line 37: We should not ignore and filter out these types of events

mukhiddin-yusuf (changes requested):

src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+
+  @Override
+  public boolean filter(ConsumerRecord<String, ResourceEvent> consumerRecord) {
+    log.info("Processing resource event [id: {}]", consumerRecord.value().getId());
+    var resourceEvent = consumerRecord.value();
+    var resourceName = resourceEvent.getResourceName();
+    if (resourceName != null) {
+      log.info("Processing resource event [resourceName: {}]", resourceName);
+      if (resourceName.equals(AUTHORITY_RESOURCE)) {
+        if (resourceEvent.getSubType() == ResourceEventSubType.HARD_DELETE) {
```
mukhiddin-yusuf, line 22: for the better code readability you can combine these 3 if statements into one.

also, I think, just to be in a safe side, it would be good to add a check for event type so that to process only ```DELETE```, as theoretically for someone it is still possible to send some other event with subtype.

mukhiddin-yusuf (changes requested):

src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+    var resourceName = resourceEvent.getResourceName();
+    var resourceEventType = resourceEvent.getType();
+    if (resourceEventType == ResourceEventType.DELETE) {
+      log.info("Processing resource event [resourceName: {}]", resourceName);
+      if (resourceName != null && resourceName.equals(AUTHORITY_RESOURCE)
+        && resourceEvent.getSubType() == ResourceEventSubType.HARD_DELETE) {
+        log.info("Skip event. No need to process event for authority resource");
+        return true;
+      }
+    }
```
psmagin, line 28: You can invert your if statements to be more safe.
```suggestion
    if (ResourceEventType.DELETE == resourceEventType) {
      log.info("Processing resource event [resourceName: {}]", resourceName);
      if (AUTHORITY_RESOURCE.equals(resourceName)
        && ResourceEventSubType.HARD_DELETE == resourceEvent.getSubType()) {
        log.info("Skip event. No need to process event for authority resource");
        return true;
      }
    }
```

src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+  public boolean filter(ConsumerRecord<String, ResourceEvent> consumerRecord) {
+    log.info("Processing resource event [id: {}]", consumerRecord.value().getId());
+    var resourceEvent = consumerRecord.value();
+    var resourceName = resourceEvent.getResourceName();
+    var resourceEventType = resourceEvent.getType();
+    if (resourceEventType == ResourceEventType.DELETE) {
+      log.info("Processing resource event [resourceName: {}]", resourceName);
+      if (resourceName != null && resourceName.equals(AUTHORITY_RESOURCE)
+        && resourceEvent.getSubType() == ResourceEventSubType.HARD_DELETE) {
+        log.info("Skip event. No need to process event for authority resource");
```
psmagin, line 25:
```suggestion
        log.info("Skip hard-delete event. No need to process event for authority resource");
```

psmagin (dismissed):

psmagin:
Also please update NEWS file

src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
psmagin: And yes, you can combine all condition in one if.

src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+public class ResourceChangeFilterStrategy implements RecordFilterStrategy<String, ResourceEvent> {
+
+  @Override
+  public boolean filter(ConsumerRecord<String, ResourceEvent> consumerRecord) {
+    log.info("Processing resource event [id: {}]", consumerRecord.value().getId());
+    var resourceEvent = consumerRecord.value();
+    var resourceName = resourceEvent.getResourceName();
+    var resourceEventType = resourceEvent.getType();
+    if (resourceEventType == ResourceEventType.DELETE) {
+      log.info("Processing resource event [resourceName: {}]", resourceName);
```
psmagin, line 22: This message if overhead. Please remove it

src/main/resources/swagger.api/schemas/resourceEvent.json
```diff
@@ -11,6 +11,10 @@
       "description": "Resource event operation type",
       "$ref": "resourceEventType.json"
     },
+    "subType": {
```
mukhiddin-yusuf, line 14: based on the field name, ```subType```, I would assume that this field is related to any resource and event type. However, in the description we are telling that this is for ```delete operation type```.

I might consider to change the description to something like ```Resource event operation sub type``` or field name to ```deleteEventSubType```. Then, similarly the corresponding changes for the schema: if we want to leave a possibility to use this field also for other event types in the future then I would change description in ```resourceEventSubType.json``` to ```Resource event operation sub type``` or the schema's name to ```resourceDeleteEventSubType``` (in case if it is just for delete event)

src/main/java/org/folio/search/integration/ResourceChangeFilterStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+import org.folio.search.domain.dto.ResourceEvent;
+import org.folio.search.domain.dto.ResourceEventType;
+import org.springframework.kafka.listener.adapter.RecordFilterStrategy;
+
+@Log4j2
+public class ResourceChangeFilterStrategy implements RecordFilterStrategy<String, ResourceEvent> {
+
+  @Override
+  public boolean filter(ConsumerRecord<String, ResourceEvent> consumerRecord) {
+    log.info("Processing resource event [id: {}]", consumerRecord.value().getId());
```
psmagin, line 17:
```suggestion
    log.debug("Processing resource event [id: {}]", consumerRecord.value().getId());
```

psmagin (commented):
Please fix code smells provided by Sonar

mukhiddin-yusuf (approved):

TsaghikKhachatryan (approved):

PR #464, opened 2023-11-07
fix(call-numbers): add null check for effectiveCallNumberComponents

TsaghikKhachatryan:
Closes: MSEARCH-622

## Purpose
Add a null check for the item.effectiveCallNumberComponents 

## Approach
Used Optional.OfNullable 

### Learning
https://issues.folio.org/browse/MSEARCH-622

### DIFF
--- a/src/main/java/org/folio/search/utils/CallNumberUtils.java
+++ b/src/main/java/org/folio/search/utils/CallNumberUtils.java
@@ -19,6 +19,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.folio.search.domain.dto.CallNumberBrowseItem;
 import org.folio.search.domain.dto.Instance;
 import org.folio.search.domain.dto.Item;
+import org.folio.search.domain.dto.ItemEffectiveCallNumberComponents;
 import org.folio.search.model.service.BrowseContext;
 import org.folio.search.model.types.CallNumberType;
 import org.folio.search.service.consortium.ConsortiumSearchHelper;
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/search/utils/CallNumberUtilsTest.java
```diff
@@ -133,6 +133,18 @@ void excludeIrrelevantResultItems(String callNumberType, List<CallNumberBrowseIt
     assertThat(unchangedItems).isEqualTo(given);
   }
 
+  @Test
+  public void excludeIrrelevantResultItems_with_null_iecnc() {
+    var context = BrowseContext.builder().build();
+    var givenItems = createItemWithNullEffectiveCallNumberComponents();
+    var callNumberTypeValue = "Z669.R360 197";
```
viacheslavkol, line 140: "callNumberTypeValue" is supposed to be a call number type name, such as "dewey"/"lc"...

src/test/java/org/folio/search/utils/CallNumberUtilsTest.java
```diff
@@ -133,6 +133,18 @@ void excludeIrrelevantResultItems(String callNumberType, List<CallNumberBrowseIt
...
+
+    var excludedItems = CallNumberUtils
```
viacheslavkol, line 142: This method returns items that remain. Invalid get filtered out and not returned

viacheslavkol (changes requested):

src/main/java/org/folio/search/utils/CallNumberUtils.java
```diff
@@ -202,8 +202,11 @@ public static List<CallNumberBrowseItem> excludeIrrelevantResultItems(BrowseCont
...
           return instance.getItems()
             .stream()
-            .anyMatch(i -> r.getFullCallNumber() == null
-              || getFullCallNumber(i).equals(r.getFullCallNumber()));
+            .anyMatch(i -> {
+              String fullCallNumber = getFullCallNumber(i);
+              return r.getFullCallNumber() == null
+                || fullCallNumber != null && fullCallNumber.equals(r.getFullCallNumber());
+            });
         }
```
mukhiddin-yusuf, line 210: looks like, from the point of better readability it makes sense to inline this block of code served as a predicate in ```.filter(...)``` to some method

mukhiddin-yusuf (changes requested):

mukhiddin-yusuf (approved):

PR #475, opened 2023-11-16
fix(cn-browse): Return result only for desired cn type

viacheslavkol:
### Purpose
Return result only for desired cn type. Return empty result with shelf key for passed cn type or normalized call number when no exact match.

### Approach
- Browse only for shelving order using desired cn type
- Browse for multiple anchors (shelf keys generated using different algorithms) only if cn type not passed

### Changes Checklist
- [ ] **API Changes**: List any API paths, methods, request or response bodies changed, added, or removed.
- [ ] **Database Schema Changes**: Specify any database schema changes and their impact.
- [ ] **Interface Dependencies**: Note added or removed dependencies.
- [ ] **NEWS**: Ensure that the `NEWS` file is updated with relevant information about the changes made in this pull request.

### Related Issues
[MSEARCH-605](https://issues.folio.org/browse/MSEARCH-605)

### Learning
Test case scenario from a task showed a case when browsing for not typed call numbers and empty result returned with shelf key generated using LcCallNumber algorithm and browse result itself was returned for browsing on same shelf key in elasticsearch which is incorrect in a case when no cn is passed. Now when call number is passed we will always look for exact match using one algorithm. When call number type is not passed - we will browse using all cn types and in case of missing the exact match - return browse result for "normalized call number" browse.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,7 @@
 * Fix secure setup of system users by default ([MSEARCH-608](https://issues.folio.org/browse/MSEARCH-608))
 * Fix result filtering for items.effectiveLocationId ([MSEARCH-615](https://issues.folio.org/browse/MSEARCH-615))
 * Ignore authority shadow copies while indexing ([MSEARCH-638](https://issues.folio.org/browse/MSEARCH-638))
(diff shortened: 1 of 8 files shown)


psmagin (approved):
NEWS please

TsaghikKhachatryan (approved):

PR #514, opened 2024-02-08
feat(browse-classification): prepare and populate database for classification browse

psmagin:
### Purpose
Prepare and populate database for classification browse

### Approach
- prepare database scripts for new table creation
- create pre-processor for instances to populdate the table with classifications (it should be used then to convert database records to elastic documents)
- replace folio-service-tools wtih folio-spring-testing that causes a lot of import changes.

### Changes Checklist
- [x] **Database Schema Changes**: Indicate any database schema changes and their impact. Confirm that migration scripts were created.
- [x] **Logging**: Confirm that logging is appropriately handled.

### Related Issues
https://folio-org.atlassian.net/browse/MSEARCH-667


mukhiddin-yusuf (approved):

src/main/java/org/folio/search/repository/classification/InstanceClassificationJdbcRepository.java
```diff
@@ -0,0 +1,164 @@
...
+        json_agg(json_build_object(
+            'instanceId', t1.instance_id,
+            'shared', t1.shared,
+            'tenantId', t1.tenant_id
+        )) AS instances
+    FROM %1$s t1
+    INNER JOIN %1$s t2 ON t1.classification_number = t2.classification_number
+                                        AND t1.classification_type = t2.classification_type
+                                        AND t2.instance_id IN (%2$s)
+    GROUP BY t1.classification_number, t1.classification_type;
```
viacheslavkol, line 47: Did you check it with "EXPLAIN ANALYZE"? Will it filter by instanceId first to prevent building aggregations for all before filtering?

src/main/java/org/folio/search/repository/classification/InstanceClassificationJdbcRepository.java
```diff
@@ -0,0 +1,164 @@
...
+    """;
+  private static final String INSERT_SQL = """
+    INSERT INTO %s (classification_type, classification_number, tenant_id, instance_id, shared)
+    VALUES (?, ?, ?, ?, ?)
+    ON CONFLICT (classification_type, classification_number, tenant_id, instance_id)
+    DO UPDATE SET shared = EXCLUDED.shared;
```
viacheslavkol, line 53: Shouldn't we also update type as it is nullable and I guess could be updated?

src/main/java/org/folio/search/repository/classification/InstanceClassificationJdbcRepository.java
viacheslavkol: And do we need filter by instanceId at all?
As I remember - we need to construct and replace full documents by type+number which will have a list of objects for each instanceId it has.
I don't see it being used for now, but I assume this is the query for constructing an elasticseach document.

src/main/java/org/folio/search/repository/classification/InstanceClassificationJdbcRepository.java
viacheslavkol: If type could be updated - then I guess we would need to left a note to delete old document before inserting the new one as id changes, in elasticsearch implementation story

src/main/java/org/folio/search/repository/classification/InstanceClassificationJdbcRepository.java
```diff
@@ -0,0 +1,164 @@
...
+      }
+      return new InstanceClassificationEntityAgg(type, number, instanceSubResources);
+    };
+  }
+
+  private String getTableName() {
+    return JdbcUtils.getFullTableName(context, INSTANCE_CLASSIFICATION_TABLE_NAME);
+  }
+
+  private String itemTypeToDatabaseValue(InstanceClassificationEntity.Id id) {
```
viacheslavkol, line 161:
```suggestion
  private String classificationTypeToDatabaseValue(InstanceClassificationEntity.Id id) {
```

src/main/resources/changelog/changes/v3.2/create_instance_classification_table.xml
```diff
@@ -0,0 +1,63 @@
...
+      <column name="classification_type" type="VARCHAR(255)">
+        <constraints primaryKey="true" primaryKeyName="pk_instance_classification"/>
+      </column>
+      <column name="classification_number" type="VARCHAR(255)">
+        <constraints primaryKey="true" primaryKeyName="pk_instance_classification"/>
+      </column>
+      <column name="tenant_id" type="VARCHAR(255)">
+        <constraints primaryKey="true" primaryKeyName="pk_instance_classification"/>
+      </column>
+      <column name="instance_id" type="VARCHAR(255)">
```
viacheslavkol, line 27: Why VARCHAR(250)?

src/main/resources/changelog/changes/v3.2/create_instance_classification_table.xml
```diff
@@ -0,0 +1,63 @@
...
+    <preConditions>
+      <not>
+        <tableExists tableName="instance_classification"/>
+      </not>
+    </preConditions>
+
+    <comment>Create instance_classification table</comment>
+
+    <createTable tableName="instance_classification">
+      <column name="classification_type" type="VARCHAR(255)">
```
viacheslavkol, line 18: "classification_type_id"? And should it be VARCHAR(255)?

README.md
```diff
@@ -264,6 +264,7 @@ and [Cross-cluster replication](https://docs.aws.amazon.com/opensearch-service/l
...
 | BROWSE_CN_INTERMEDIATE_VALUES_ENABLED              | true                                                       | Specifies if globally intermediate values (nested instance items) must be populated or not (tenant can override this setting)                                                         |
 | BROWSE_CN_INTERMEDIATE_REMOVE_DUPLICATES           | true                                                       | Specifies if globally intermediate duplicate values (fullCallNumber) should be removed or not (Active only with BROWSE_CN_INTERMEDIATE_VALUES_ENABLED)                                |
+| BROWSE_CLASSIFICATIONS_ENABLED                     | false                                                      | Specifies if globally instance classification indexing will be performed                                                                                                              |
```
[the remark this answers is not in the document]
psmagin: It's false until we will have all done for classification browse. And due to the feature is at risk for current release.

src/main/java/org/folio/search/repository/classification/InstanceClassificationJdbcRepository.java
psmagin: It uses 2 indexes:
image
image


As indexing of classification is part of instances indexing and preparing documents during instances processing, I think it's one of the way to find all classifications related to instances that currently in process.
This query takes all classifications that instances have and aggregate them in elastic-like document, so with this quesry we will have full document.

src/main/java/org/folio/search/repository/classification/InstanceClassificationJdbcRepository.java
psmagin: classification_type is part ok PK for identifing uniqueness of classification. From perspective of UI classification_number is also updatable. In case of some of these fields updated - it's identified as new classification and old one with previous values will be deleted.

src/main/resources/changelog/changes/v3.2/create_instance_classification_table.xml
psmagin: We can reduce VARCHAR size to UUID-size but it can't be changed to UUID type. As this field is nullable from business logic but can't be null in database as it is part of PK, I use `<null>` replacement for it.

src/main/resources/changelog/changes/v3.2/create_instance_classification_table.xml
psmagin: It can be reduced to size of UUID. Not using of UUID type caused by wish to reduce amount of operations (converting UUID to String, String to UUID in java code and casting text to uuid on database side) and following approach of consortium_instance table.

TsaghikKhachatryan (approved):

PR #522, opened 2024-02-24
feat(browse-classifications): implement endpoint to browse by classifications

psmagin:
### Purpose
Implement endpoint to browse by classifications

### Approach
- Define API
- Implement new endpoint logic
-- Change logic of indexing to create shelving order for each algorithm
-- Use browse config to find right field for browse
-- Moved instance pre-processor earlier in a flow due to consortium complication logic

### Changes Checklist
- [x] **API Changes**: Document any API paths, methods, request or response bodies changed, added, or removed.
- [x] **Interface Version Changes**: Indicate any changes to interface versions.
- [x] **Permissions**: Document any changes to permissions.
- [x] **Logging**: Confirm that logging is appropriately handled.

### Related Issues
https://folio-org.atlassian.net/browse/MSEARCH-665


GeloPakDev1 (approved):

src/main/java/org/folio/search/service/browse/ClassificationBrowseService.java
```diff
@@ -0,0 +1,136 @@
...
+      if (anchorQuery != null) {
+        return anchorQuery;
+      }
+      return matchAllQuery();
+    } else {
+      var boolQueryMain = boolQuery();
+      if (!config.getTypeIds().isEmpty()) {
+        var boolQuery = boolQuery();
+        for (var typeId : typeIds) {
+          boolQuery.should(termQuery(CLASSIFICATION_TYPE_ID_FIELD, typeId.toString()));
```
viacheslavkol, line 109: You can add "should" clauses to a main query if you have "minimumShouldMatch(1)" in it. Unless there's any other reason to use such query nesting

src/main/java/org/folio/search/utils/ShelvingOrderCalculationHelper.java
```diff
@@ -0,0 +1,19 @@
...
+import org.marc4j.callnum.LCCallNumber;
+
+@UtilityClass
+public class ShelvingOrderCalculationHelper {
+
+  public static String calculate(@NonNull String input, @NonNull ShelvingOrderAlgorithmType algorithmType) {
+    return switch (algorithmType) {
+      case LC -> new LCCallNumber(input).getShelfKey().trim();
+      case DEWEY -> new DeweyCallNumber(input).getShelfKey().trim();
+      case DEFAULT -> CallNumberUtils.normalizeEffectiveShelvingOrder(input);
```
viacheslavkol, line 16: I don't think this is acceptable to use as internally it has "isSupportedCharacter" which will replace characters not supporter by call number numeric algorithm

viacheslavkol (changes requested):

src/main/java/org/folio/search/service/browse/ClassificationBrowseService.java
psmagin: It's look more clear to me.
And also I think that in consortium there will be added additional active affiliation should queries that could break filtering logic.

PR #542, opened 2024-03-19
MSEARCH-696: Make all searches case-insensitive

TsaghikKhachatryan:
### Purpose
Make Inventory and Authority searches case-insensitive.

### Approach
Change the field types of "keyword" to "keyword_lowercase".

Create a custom analyzer named "whitespace_analyzer" with the "lowercase" filter and add it as an analyzer to the existing "whitespace" indices.

### Changes Checklist
- [ ] **API Changes**: Document any API paths, methods, request, or response bodies changed, added, or removed.

### Related Issues
https://folio-org.atlassian.net/browse/MSEARCH-696
https://folio-org.atlassian.net/browse/MSEARCH-688


src/main/java/org/folio/search/utils/SearchUtils.java
```diff
@@ -307,6 +307,16 @@ public static boolean isEmptyString(Object value) {
     return value instanceof String string && string.isEmpty();
   }
 
+  /**
+   * Check if passed value is blank.
+   *
+   * @param value - value to check
+   * @return true - if value is blank, false - otherwise
+   */
+  private static boolean isBlankString(String value) {
```
psmagin, line 316: I don't think that this method makes sense, as there no other logic apart from using string utils method

src/main/resources/elasticsearch/index/instance.json
```diff
@@ -33,6 +33,11 @@
         "filter": [ "folio_word_delimiter_graph", "icu_folding" ],
         "char_filter": [ "and_char_replacement" ],
         "type": "custom"
+      },
+      "whitespace_analyzer": {
```
psmagin, line 37: whitespace_lowercase probably will make it consistent with other

src/main/resources/model/authority.json
```diff
@@ -7,11 +7,11 @@
   },
   "fields": {
     "id": {
-      "index": "keyword",
+      "index": "keyword_lowercase",
       "showInResponse": [ "search", "browse" ]
     },
     "tenantId": {
-      "index": "keyword",
+      "index": "keyword_lowercase",
```
psmagin, line 14: Please back it to just keyword as well as other tenantId fields

src/main/resources/model/authority.json
```diff
@@ -323,8 +323,9 @@
       "default": "NULL"
     },
     "naturalId": {
-      "index": "keyword",
-      "showInResponse": [ "search", "browse" ]
+      "index": "keyword_lowercase",
+      "showInResponse": [ "search", "browse" ],
+      "searchTermProcessor": "naturalIdSearchTermProcessor"
     },
```
psmagin, line 329: Why there a need for processor that makes value lowecase if we already use keyword-lowercase?

psmagin (changes requested):

src/main/resources/elasticsearch/index/instance.json
TsaghikKhachatryan: Are you asking if it is consistent with  keyword_lowercase? I guess, that the distinction is that "whitespace" enables full-text searches, whereas "keyword" makes exact matches. Originally, "whitespace" was used for identifiers.value. Therefore, I included an additional analyzer to preserve the original functionality

psmagin: I mean probably will be better to rename it to whitespace_lowercase_analyzer

psmagin (approved):

viacheslavpoliakov (approved):

PR #573, opened 2024-04-26
fix(facets): Fix facets for local call number types

viacheslavkol:
### Purpose
Fix call number type facet filter.

### Approach
- Fetch reference data for "folio" source call number types instead of local to determine if type is local
Reference data is cached for one hour so when new local call number type is created it will be absent in cache. Using folio call number types eliminates this problem as folio types couldn't be created from a UI.

### Changes Checklist

### Related Issues
[MSEARCH-738](https://folio-org.atlassian.net/browse/MSEARCH-738)

### DIFF
--- a/src/test/java/org/folio/search/controller/BrowseCallNumberConsortiumIT.java
+++ b/src/test/java/org/folio/search/controller/BrowseCallNumberConsortiumIT.java
@@ -62,9 +62,8 @@ class BrowseCallNumberConsortiumIT extends BaseConsortiumIntegrationTest {
     setUpTenant(CENTRAL_TENANT_ID, INSTANCES_CENTRAL);
     setUpTenant(MEMBER_TENANT_ID);
     saveRecords(MEMBER_TENANT_ID, instanceSearchPath(), asList(INSTANCES_MEMBER),
-      INSTANCES_CENTRAL.length,
+      4,
       instance -> inventoryApi.createInstance(MEMBER_TENANT_ID, instance));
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/search/controller/BrowseCallNumberConsortiumIT.java
```diff
@@ -0,0 +1,169 @@
...
+  private static final Instance[] INSTANCES_CENTRAL = instancesCentral();
+  private static final Instance[] INSTANCES_MEMBER = instancesMember();
+
+
+  @BeforeAll
+  static void prepare(@Autowired RestHighLevelClient restHighLevelClient) throws InterruptedException {
+    setUpTenant(CENTRAL_TENANT_ID, INSTANCES_CENTRAL);
+    setUpTenant(MEMBER_TENANT_ID);
+    saveRecords(MEMBER_TENANT_ID, instanceSearchPath(), asList(INSTANCES_MEMBER),
+      INSTANCES_CENTRAL.length,
```
mukhiddin-yusuf, line 65: maybe you wanted to put here: ```INSTANCES_MEMBER``` or I did not get the idea - saving member's instances but expect 2 events being received ?

mukhiddin-yusuf: also, why in a next line to save the member's instances again?

viacheslavkol: Yes, typo, changed to 4 expected instances.
Copied from some old local test class, didn't notice, thanks, removed

mukhiddin-yusuf (approved):

viacheslavpoliakov (approved):

PR #597, opened 2024-06-04
MSEARCH-782: Fix count for Shared facet in Contributor and Subjects browse

mukhiddin-yusuf:
### Purpose
_Shared facet for Contributors and Subjects browse show incorrect values of shared/non-shared instances_

### Approach
_Handle update event for instance sharing appropriately_

### Changes Checklist

### Related Issues
[MSEARCH-782](https://folio-org.atlassian.net/browse/MSEARCH-782)


src/main/java/org/folio/search/integration/KafkaMessageProducer.java
```diff
@@ -75,8 +77,20 @@ private void prepareAndSendEvents(List<ResourceEvent> resourceEvents,
...
+    var shared = isSharedResource(tenantId);
+    var oldSubjects = extractSubjects(getOldAsMap(event), shared);
+    var newSubjects = extractSubjects(getNewAsMap(event), shared);
+
+    if (isUpdateEventForResourceSharing(event)) {
+      if (Boolean.TRUE.equals(shared)) {
+        log.warn("Update event for instance sharing is supposed to be for member tenant,"
+          + " but received for central: {}", tenantId);
+      }
+      return prepareSubjectsEvents(oldSubjects, List.of(), tenantId, DELETE);
```
viacheslavkol, line 89: Why only delete?

mukhiddin-yusuf: After instance is shared there are two records of subjects/contributors for the same instance in the index for subjects/contributors: one for the member which was created before instance is shared (and has shared=false) and one for the central which is added after sharing (and has shared=true). Here I've added the logic for handling the update event received for instance sharing, so here we need to delete the record in the index for member tenant (with shared=false), and keep the one for the central tenant. Thus we only need delete event here.

psmagin (approved):

viacheslavpoliakov (approved):

PR #612, opened 2024-06-27
MSEARCH-785: Increase limit for batch request IDs size in search of Items/Holdings in consortium

mukhiddin-yusuf:
### Purpose
_Increase batch IDs limit for search consolidated items/holdings in consortium_

### Approach
_modify elasticsearch query and use terms query_

### Changes Checklist

### Related Issues
[MSEARCH-785](https://folio-org.atlassian.net/browse/MSEARCH-785)


src/main/java/org/folio/search/service/consortium/ConsortiumInstanceSearchService.java
```diff
@@ -67,34 +89,105 @@ public ConsortiumItem getConsortiumItem(String id, CqlSearchRequest<Instance> se
...
+      return List.of(SearchResult.of(records.size(), records));
+    }
+
+    var searchSourceBuilder = queryBuilder(termsQuery, properties.getSearchConsortiumRecordsPageSize())
+      .sort(fieldSort(targetField).order(ASC))
+      .searchAfter(new Object[]{""});
+    var response = searchRepository.search(request, searchSourceBuilder);
+    List<SearchResult<T>> searchRecords = new ArrayList<>();
+
+    while (response.getHits() != null && response.getHits().getHits().length > 0) {
```
viacheslavkol, line 148: Can't we use `searchRepository.msearch` instead, by partitioning `ids` using `DEFAULT_MAX_SEARCH_RESULT_WINDOW`?

src/main/java/org/folio/search/service/consortium/ConsortiumInstanceSearchService.java
```diff
@@ -67,34 +89,105 @@ public ConsortiumItem getConsortiumItem(String id, CqlSearchRequest<Instance> se
...
+      var response = searchRepository.search(request, searchSourceBuilder);
+      var searchResult = documentConverter.convertToSearchResult(response, request.getResourceClass(),
+        (hits, item) -> recordMapper.apply(item, ids));
+      var records = searchResult.getRecords().stream()
+        .flatMap(List::stream)
+        .toList();
+      return List.of(SearchResult.of(records.size(), records));
+    }
+
+    var searchSourceBuilder = queryBuilder(termsQuery, properties.getSearchConsortiumRecordsPageSize())
```
viacheslavkol, line 142: Why are we querying 5_000 (default) here, while above we can query more while it's under 10_000, even 9_000?

viacheslavkol (commented):
Fix sonar issues, please

src/main/java/org/folio/search/service/consortium/ConsortiumInstanceSearchService.java
mukhiddin-yusuf: technically it is possible, but I think (deep) paginating with ```search_after``` is a solution tailored for this kind of cases. When we partition ```ids``` and run multiple query each query will be separate one, but with ```search_after``` each query will have necessary information to retrieve the next page of results.

mukhiddin-yusuf: actually, here we are not dealing with an issue of deep pagination where there may be case of skipping over ```from``` number of hits and thus the need to keep in memory those documents, we are just overcoming the constraint of ```index.max_result_window``` which is by default set as 10_000. So, maybe partitioning the ```ids``` and running multiple queries (possibly with parallelism) would be faster, but I do not have any result to compare and say something now. After the merge I can test this approach on environment with more data and assess the run times.

viacheslavkol (approved):

psmagin (approved):

viacheslavpoliakov (approved):

PR #633, opened 2024-07-23
[MSEARCH-812] Increase the limit to fetch more than 1k locations

azizbekxm:
### Purpose
In scope of ticket we need to fetch all available locations to check by code https://github.com/folio-org/mod-gobi/pull/238
### Approach
Increasing max limit for api search/consortium/locations
https://folio-org.atlassian.net/browse/MSEARCH-812

### DIFF
--- a/src/main/java/org/folio/search/service/consortium/ConsortiumLocationService.java
+++ b/src/main/java/org/folio/search/service/consortium/ConsortiumLocationService.java
@@ -45,10 +45,9 @@ public class ConsortiumLocationService {
 
   private void validatePaginationParameters(Integer limit, Integer offset) {
     if (limit + offset > MAX_RESULT_WINDOW) {
-      throw new IllegalArgumentException("The combination of limit and offset exceeds " +
-        "the maximum result window of" + MAX_RESULT_WINDOW);
+      throw new IllegalArgumentException("The combination of limit and offset exceeds "
+        + "the maximum result window of" + MAX_RESULT_WINDOW);
     }
   }
-
 
 }
--- a/src/main/java/org/folio/search/service/consortium/ConsortiumLocationService.java
+++ b/src/main/java/org/folio/search/service/consortium/ConsortiumLocationService.java
@@ -1,9 +1,12 @@
 package org.folio.search.service.consortium;
 
+import static org.folio.search.service.SearchService.DEFAULT_MAX_SEARCH_RESULT_WINDOW;
+
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 import org.folio.search.domain.dto.ConsortiumLocation;
 import org.folio.search.domain.dto.SortOrder;
+import org.folio.search.exception.RequestValidationException;
 import org.folio.search.model.SearchResult;
 import org.folio.search.repository.ConsortiumLocationRepository;
 import org.springframework.stereotype.Service;
@@ -16,7 +19,6 @@ public class ConsortiumLocationService {
   public static final String NAME = "name";
   public static final String ID = "id";
   public static final String TENANT_ID = "tenantId";
-  private static final int MAX_RESULT_WINDOW = 10000;
   private final ConsortiumLocationRepository repository;
   private final ConsortiumTenantExecutor executor;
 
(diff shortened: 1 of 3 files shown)


SerhiiNosko (approved):

src/main/resources/swagger.api/parameters/consortium-locations-limit-param.yaml
```diff
@@ -4,5 +4,5 @@ description: Limit the number of elements returned in the response.
 schema:
   type: integer
   minimum: 0
-  maximum: 1000
+  maximum: 10000
```
mukhiddin-yusuf, line 7: keep in mind that this API uses pagination search and opensearch/elasticsearch puts a limit of 10.000 search results for ```from + size``` (or ```offset + limit```)
https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html
https://opensearch.org/docs/latest/search-plugins/searching-data/paginate/

now, if API is called with max limit then offset must be 0, or if the use of using this API with ```offset + limit > 10.000``` is possible or quite frequent then search query needs to be changed

SerhiiNosko: we will use this without limit and offset param, fetching only 1000 can not be enought, but 10k is good. So 10000 as ES suggests as offset + limit

psmagin: I suggest to add validation in code to check that offset + limit is less then 10000

psmagin:
Will the story https://folio-org.atlassian.net/browse/MSEARCH-809 be closed after this PR merge? If no, then I suggest to split it

src/main/resources/swagger.api/parameters/consortium-locations-limit-param.yaml
SerhiiNosko: added AC regarding this into newly created ticket: https://folio-org.atlassian.net/browse/MSEARCH-812

SerhiiNosko:
> Will the story https://folio-org.atlassian.net/browse/MSEARCH-809 be closed after this PR merge? If no, then I suggest to split it

Extracted separate story for this: https://folio-org.atlassian.net/browse/MSEARCH-812
By the way, we checked that /search/consortium/locations is working also in non-ECS mode so we are going to use it in both ECS and non-ECS for mod-gobi to avoid an additional check if ECS mode is enabled. If you see some issues in this approach please let us know.

src/main/resources/swagger.api/parameters/consortium-locations-limit-param.yaml
azizbekxm: Added validation for this case

src/main/java/org/folio/search/service/consortium/ConsortiumLocationService.java
```diff
@@ -41,5 +43,11 @@ private void validateSortByValue(String sortBy) {
     }
   }
 
+  private void validatePaginationParameters(Integer limit, Integer offset) {
+    if (limit + offset > MAX_RESULT_WINDOW) {
+      throw new IllegalArgumentException("The combination of limit and offset exceeds "
+        + "the maximum result window of" + MAX_RESULT_WINDOW);
```
psmagin, line 49:
```suggestion
        + "the maximum result window of " + MAX_RESULT_WINDOW);
```

src/test/java/org/folio/search/service/consortium/ConsortiumLocationServiceTest.java
```diff
@@ -59,6 +60,29 @@ void fetchLocations_ValidSortBy() {
     verify(executor).execute(eq(tenantId), any(Supplier.class));
   }
 
+
+  @Test
+  void fetchCampuses_InvalidSortBy() {
```
psmagin, line 65:
```suggestion
  void fetchLocations_InvalidSortBy() {
```

src/test/java/org/folio/search/service/consortium/ConsortiumLocationServiceTest.java
```diff
@@ -59,6 +60,29 @@ void fetchLocations_ValidSortBy() {
...
+    var limit = 10;
+    var offset = 0;
+
+    Assertions.assertThrows(IllegalArgumentException.class, () ->
+      service.fetchLocations(CONSORTIUM_TENANT, CONSORTIUM_TENANT, limit, offset, "invalid", sortOrder)
+    );
+  }
+
+  @Test
+  void fetchCampuses_InvalidPaginationParameters() {
```
psmagin, line 76:
```suggestion
  void fetchLocations_InvalidPaginationParameters() {
```

src/main/java/org/folio/search/service/consortium/ConsortiumLocationService.java
azizbekxm: thanks fixed

src/test/java/org/folio/search/service/consortium/ConsortiumLocationServiceTest.java
azizbekxm: thanks, fixed

src/test/java/org/folio/search/service/consortium/ConsortiumLocationServiceTest.java
azizbekxm: fixed as well

mukhiddin-yusuf (approved):

PR #646, opened 2024-08-22
feat(instance-search): create computed field for sorting and filtering Date1

SvitlanaKovalova1:
### Purpose
[MSEARCH-806](https://folio-org.atlassian.net/browse/MSEARCH-806) Currently in the Instance schema, there is a field for publication date. However, this data originates from an uncontrolled string in variable fields of the MARC record. The dates in the 008 are more consistently formatted, and are more useful for certain materials like serials. Previously, we implemented sortable properties (publicationPeriod.start/publicationPeriod.end) that are parsed from the publication date field. We need to break this relationship, and instead make the source of that data the 07-14 from the 008. The 008 structure allows for a single position to indicate the type of date that will be represented in the following positions.
008/06 = Date type
008/07-10 = Date 1
008/11-14 = Date 2

For this story, we need to create computed fields from Date 1 for sorting and filtering dates. And we need validation on the backend that stores those validation rules, and if the values in Date 1 don’t pass validation, we should consider as zeroes so that they still show in results list and show in filters

### Requirements:
- If Date 1 contains “u” in one of the four characters, replace “u” with 0
- If no data, consider as 0
- Store validation rules for Date 1:
      - Each can only store up to four characters
      - Allows for numeric values and the alpha “u”
- If characters in Date 1 do not pass this validation, consider the values as 0
- Return Date 1, Date 2, and Date type
- Verify the ability to search original data (not computed field) in “All” search; example queries:
      - Support trailing, leading, and internal wildcards and exact matches (15*, *02, 19*7, 1970)
      - Dates with “u” should follow the original data (not the computed field with “u” replaced with “0”): 199u exact match, would also match “199*”

### Approach
Update index:
1. Add new Instance schema fields: dateTypeId, date 1, date 2. It is searchable based on requirements.
2. Add a new searchable field: sort_date1.  It is computed from date 1 based on requirements. It  works with range queries and is used for sorting.

### Changes Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,7 +561,7 @@ does not produce any values, so the following search options will return an empt
 | `normalizedClassificationNumber`       |   term    | `normalizedClassificationNumber == "LCCN"`                           | Matches instances with the given classification number (normalizes case, whitespaces, special characters, supports leading and trailing wildcard) |
 | `dates.date1`                          |   term    | `dates.date1="199*"`                                                 | Matches instances with the given Date1 (supports leading, trailing  and internal wildcards)                                                       |
 | `dates.date2`                          |   term    | `dates.date2="199*"`                                                 | Matches instances with the given Date2 (supports leading, trailing  and internal wildcards)                                                       |
-| `sort_date1`                           |   term    | `sort_date1=199u`                                                    | Matches instances with the given Date1 (normalizes  alpha 'u' characters, supports leading, trailing  and internal wildcards)                     |
+| `sort_date1`                           |   term    | `sort_date1=1990`                                                    | Matches instances with the given Date1 (normalizes  alpha 'u' characters, supports leading, trailing  and internal wildcards)                     |
 
 ##### Holdings search options
 
--- a/README.md
+++ b/README.md
@@ -541,7 +541,7 @@ does not produce any values, so the following search options will return an empt
 | `modeOfIssuanceId`                     |   term    | `modeOfIssuanceId=="123"`                                            | Matches instances that have `123` mode of issuance                                                                                                |
 | `natureOfContentTermIds`               |   term    | `natureOfContentTermIds=="123"`                                      | Matches instances that have `123` nature of content                                                                                               |
 | `publisher`                            | full-text | `publisher all "Publisher of Ukraine"`                               | Matches instances that have `Publisher of Ukraine` publisher                                                                                      |
-| `publication.place`                    | full-text   | `publication.place all "Ukraine"`                                       | Matches instances that have `Ukraine` in publication place                                                                                           |
+| `publication.place`                    | full-text   | `publication.place all "Ukraine"`                                    | Matches instances that have `Ukraine` in publication place                                                                                        |
 | `instanceTags`                         |   term    | `instanceTags=="important"`                                          | Matches instances that have `important` tag                                                                                                       |
 | `classifications.classificationNumber` |   term    | `classifications.classificationNumber=="cl1"`                        | Matches instances that have `cl1` classification number                                                                                           |
 | `classifications.classificationTypeId` |   term    | `classifications.classificationTypeId=="123"`                        | Matches instances that have classification type id `123`                                                                                          |
@@ -561,7 +561,7 @@ does not produce any values, so the following search options will return an empt
 | `normalizedClassificationNumber`       |   term    | `normalizedClassificationNumber == "LCCN"`                           | Matches instances with the given classification number (normalizes case, whitespaces, special characters, supports leading and trailing wildcard) |
 | `dates.date1`                          |   term    | `dates.date1="199*"`                                                 | Matches instances with the given Date1 (supports leading, trailing  and internal wildcards)                                                       |
 | `dates.date2`                          |   term    | `dates.date2="199*"`                                                 | Matches instances with the given Date2 (supports leading, trailing  and internal wildcards)                                                       |
-| `sort_date1`                           |   term    | `sort_date1=1990`                                                    | Matches instances with the given Date1 (normalizes  alpha 'u' characters, supports leading, trailing  and internal wildcards)                     |
+| `normalizedDate1`                      |   term    | `normalizedDate1>=1990`                                              | Matches instances with the given Date1 (normalizes  alpha 'u' characters)                                                                         |
 
 ##### Holdings search options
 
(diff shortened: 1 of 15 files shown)


src/main/resources/model/instance.json
```diff
@@ -127,6 +127,23 @@
...
+          "showInResponse": [ "search"]
+        },
+        "date1": {
+          "index": "keyword_lowercase",
+          "showInResponse": [ "search" ]
+        },
+        "date2": {
+          "index": "keyword_lowercase",
+          "showInResponse": [ "search" ]
+        }
```
psmagin, line 144: This 2 also should be `source` as we shouldn't support search directly by them.

README.md
```diff
@@ -559,6 +559,9 @@ does not produce any values, so the following search options will return an empt
...
 | `normalizedClassificationNumber`       |   term    | `normalizedClassificationNumber == "LCCN"`                           | Matches instances with the given classification number (normalizes case, whitespaces, special characters, supports leading and trailing wildcard) |
+| `dates.date1`                          |   term    | `dates.date1="199*"`                                                 | Matches instances with the given Date1 (supports leading, trailing  and internal wildcards)                                                       |
+| `dates.date2`                          |   term    | `dates.date2="199*"`                                                 | Matches instances with the given Date2 (supports leading, trailing  and internal wildcards)                                                       |
+| `sort_date1`                           |   term    | `sort_date1=199u`                                                    | Matches instances with the given Date1 (normalizes  alpha 'u' characters, supports leading, trailing  and internal wildcards)                     |
```
psmagin, line 564: Why not `normalizedDate1` ?

psmagin:
```suggestion
| `sort_date1`                           |   term    | `sort_date1=1990`                                                    | Matches instances with the given Date1 (normalizes  alpha 'u' characters, supports leading, trailing  and internal wildcards)                     |
```
As you don't do input term normalization.

src/main/resources/model/instance.json
```diff
@@ -552,6 +569,12 @@
       "index": "keyword_lowercase",
       "processor": "sortTitleProcessor"
     },
+    "sort_date1": {
+      "searchTypes": [ "filter", "sort" ],
+      "type": "search",
+      "index": "keyword_lowercase",
```
psmagin, line 575: I think we should make it an `number` type to make sure that we will have correct number-based order.

src/main/java/org/folio/search/service/setter/instance/Date1FieldProcessor.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.folio.search.domain.dto.Dates;
+import org.folio.search.domain.dto.Instance;
+import org.folio.search.service.setter.FieldProcessor;
+import org.springframework.stereotype.Component;
+
+@Component
+public class Date1FieldProcessor implements FieldProcessor<Instance, String> {
+
+  private static final Pattern FOUR_DIGIT_REGEX = Pattern.compile("^\\d{4}$");
```
psmagin, line 13: It could be less than 4 numbers, like 998, 503, 23, 1

psmagin (changes requested):

README.md
SvitlanaKovalova1: The "sort_" prefix in the field name is required for the sortBy operation. By default, the "sort_" prefix is ​​added during the sortBy operation in CqlSortProvider, and if it cannot find a field with this prefix, we get this error: "No mapping found for [sort_normalizedDate1] in order to sort on".

- field name: "normalizedDate1"
- error:
image

src/main/resources/model/instance.json
SvitlanaKovalova1: updated

src/main/java/org/folio/search/service/setter/instance/Date1FieldProcessor.java
SvitlanaKovalova1: updated

src/main/resources/model/instance.json
SvitlanaKovalova1: updated

README.md
psmagin: Could we do smth similar to authority mappings?
image

src/test/java/org/folio/search/service/setter/instance/Date1FieldProcessorTest.java
```diff
@@ -20,21 +20,23 @@ class Date1FieldProcessorTest {
...
+      arguments(instance("20u2"), Short.valueOf("2002")),
+      arguments(instance("20uu"), Short.valueOf("2000")),
+      arguments(instance("199"), Short.valueOf("199")),
+      arguments(instance("1u9"), Short.valueOf("109")),
+      arguments(instance("1"), Short.valueOf("1")),
+      arguments(instance("19999"), Short.valueOf("0")),
+      arguments(instance("19k5"), Short.valueOf("0")),
+      arguments(new Instance(), Short.valueOf("0"))
     );
   }
```
psmagin, line 41: What about leading zeros? Cases: 0999, uuu9

README.md
```diff
@@ -559,6 +559,9 @@ does not produce any values, so the following search options will return an empt
...
 | `normalizedClassificationNumber`       |   term    | `normalizedClassificationNumber == "LCCN"`                           | Matches instances with the given classification number (normalizes case, whitespaces, special characters, supports leading and trailing wildcard) |
+| `dates.date1`                          |   term    | `dates.date1="199*"`                                                 | Matches instances with the given Date1 (supports leading, trailing  and internal wildcards)                                                       |
+| `dates.date2`                          |   term    | `dates.date2="199*"`                                                 | Matches instances with the given Date2 (supports leading, trailing  and internal wildcards)                                                       |
+| `sort_date1`                           |   term    | `sort_date1=1990`                                                    | Matches instances with the given Date1 (normalizes  alpha 'u' characters, supports leading, trailing  and internal wildcards)                     |
```
psmagin, line 564: So it doesn't normalize input

src/test/java/org/folio/search/service/setter/instance/Date1FieldProcessorTest.java
psmagin: Could you add it as a test case?

psmagin (approved):
Good job!

viacheslavkol (approved):

PR #663, opened 2024-09-25
fix(search-instances): Fix backslash handling for date1/date2

viacheslavkol:
### Purpose
Fix backslash handling for date1/date2

### Approach
Change index type from source to standard

### Changes Checklist
- [x] **Database Schema Changes**: Indicate any database schema changes and their impact. Confirm that migration scripts were created.

### Related Issues
[MSEARCH-839](https://folio-org.atlassian.net/browse/MSEARCH-839)

### Learning and Resources (if applicable)
Other fields mentioned in a story that can be queried with backslashes are all fulltext types.


SvitlanaKovalova1 (approved):

src/main/resources/model/instance.json
```diff
@@ -135,11 +135,11 @@
...
         },
         "date1": {
-          "index": "source",
+          "index": "standard",
           "showInResponse": [ "search" ]
         },
         "date2": {
-          "index": "source",
+          "index": "standard",
           "showInResponse": [ "search" ]
```
psmagin, line 143: Is it required to have it as full-text type? Could we modify the All-value processor to achieve the result without changing a type from source? There is more fields that have source type, so they will have same problem.

psmagin (approved):
NEWS please

PR #670, opened 2024-10-07
MSEARCH-852: fix linked-data resources search in consortia mode and central tenant

PBobylev:
### Purpose
Fix Linked Data resources search not working in Consortia mode and central tenant

### Approach
Add "tenantId" field to LD resources. Ignore "shared" facet during search for LD resource in centralTenant.

### Changes Checklist

### Related Issues
[MSEARCH-852](https://folio-org.atlassian.net/browse/MSEARCH-852)


AndreiBordak (approved):

src/test/java/org/folio/search/controller/SearchLinkedDataWorkConsortiumIT.java
```diff
@@ -0,0 +1,32 @@
...
+@IntegrationTest
+class SearchLinkedDataWorkConsortiumIT extends BaseConsortiumIntegrationTest {
+
+  @BeforeAll
+  static void prepare() {
+    setUpTenant(LinkedDataWork.class, CENTRAL_TENANT_ID, getWorkSampleAsMap());
+  }
+
+  @Test
+  void searchByLinkedDataInstanceInCentralTenant_shouldReturnCentralTenantRecord() throws Throwable {
```
askhat-abishev, line 26: searchByLinkedData**Work**InCentralTenant_shouldReturnCentralTenantRecord

askhat-abishev (approved):

src/test/java/org/folio/search/controller/SearchLinkedDataWorkConsortiumIT.java
PBobylev: fixed, ty!

src/main/java/org/folio/search/service/consortium/ConsortiumSearchHelper.java
```diff
@@ -51,7 +51,7 @@ public QueryBuilder filterQueryForActiveAffiliation(QueryBuilder query, Resource
   public QueryBuilder filterQueryForActiveAffiliation(QueryBuilder query, ResourceType resource,
                                                       String contextTenantId) {
     var centralTenantId = consortiumTenantService.getCentralTenant(contextTenantId);
-    if (centralTenantId.isEmpty()) {
+    if (centralTenantId.isEmpty() || resource.isLinkedDataResource() && centralTenantId.get().equals(contextTenantId)) {
       return query;
     }
```
psmagin, line 56: The PR description mentions that you should ignore the `shared` filter in consortia mode. But here, I can see that it's ignored only for central tenant requests. Where is inaccuracy?

src/main/resources/model/linked_data_authority.json
```diff
@@ -6,6 +6,11 @@
     "id": {
       "index": "keyword"
     },
+    "tenantId": {
+      "index": "keyword",
+      "searchTypes": [ "facet", "filter" ],
+      "showInResponse": [ "search" ]
+    },
```
psmagin, line 13: Will the `tenantId` be the field of incoming Kafka messages bodies?
We have an ability to get tenantId from Kafka headers, you can check other index mappings for tenantId field definition, if needed.

src/test/java/org/folio/search/support/base/BaseConsortiumIntegrationTest.java
```diff
@@ -33,38 +23,6 @@ protected static void removeTenant() {
...
-  protected static void enableTenant(String tenant) {
-    var tenantAttributes = new TenantAttributes().moduleTo("mod-search");
-    tenantAttributes.addParametersItem(new Parameter("centralTenantId").value(CENTRAL_TENANT_ID));
-
-    mockMvc.perform(post("/_/tenant", randomId())
-        .content(asJsonString(tenantAttributes))
-        .headers(defaultHeaders(tenant))
-        .contentType(APPLICATION_JSON))
-      .andExpect(status().isNoContent());
-  }
```
psmagin, line 66: What is the reason for these changed?

src/main/resources/model/linked_data_authority.json
PBobylev: Yes, it will. Took this definition from the Instance model initially. But I see your point: some other models contains the "isTenant": true". Added it ti LD models. Thank you!

src/test/java/org/folio/search/support/base/BaseConsortiumIntegrationTest.java
PBobylev: Code deduplication: BaseIntegrationTest already contains the same methods or methods which closely similair.

src/main/java/org/folio/search/service/consortium/ConsortiumSearchHelper.java
PBobylev: The Linked Data module is proposed to be installed/activated in a Central tenant only, thus there is no difference for it. And this check is jsut additional safety check. But let me clarify PR description, thank you

PBobylev:
> Update NEWS

done, thank you!

psmagin (approved):

PR #680, opened 2024-10-15
feat(indexing): simplify indexing for instance child documents

psmagin:
### Purpose
Simplify indexing for instance child documents

### Approach
- Simplify structure of classification/contributor/subject indexes 
- Modify SQL for retrieving documents above
- Cleanup indexes:
  - Remove unused metadata fields
  - Do not use icu_folding filter for fields that couldn't contain any special symbols

### Changes Checklist

### Related Issues
https://folio-org.atlassian.net/browse/MSEARCH-851


src/main/java/org/folio/search/model/index/InstanceSubResource.java
```diff
@@ -1,13 +1,14 @@
...
 
 @Data
 @Builder
 public class InstanceSubResource {
-  private String instanceId;
-  private String typeId;
   private String tenantId;
   private Boolean shared;
+  private int count;
+  private List<String> typeId;
```
viacheslavkol, line 13:
```suggestion
  private List<String> typeIds;
```

viacheslavkol (approved):

src/main/java/org/folio/search/model/index/InstanceSubResource.java
psmagin: It's typeId in the index, so to not break already existed query options it should stay `typeId`

SvitlanaKovalova1 (approved):

PR #702, opened 2024-11-28
feat(cn-browse): Implement Database Structure and Logic for Managing Call Numbers

psmagin:
### Purpose
Implement Database Structure and Logic for Managing Call Numbers

### Approach
- add new database tables
- create logic for extracting and saving call-numbers and their relations
- create a new feature flag for call-number browse.

### Changes Checklist

### Related Issues
https://folio-org.atlassian.net/browse/MSEARCH-862


src/main/java/org/folio/search/service/converter/preprocessor/extractor/impl/CallNumberResourceExtractor.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Override
+  protected List<Map<String, Object>> constructRelations(boolean shared, ResourceEvent event,
+                                                         List<Map<String, Object>> entities) {
+    return entities.stream()
+      .map(entity -> InstanceCallNumberEntity.builder()
+        .callNumberId(getString(entity, "id"))
+        .itemId(getString(getNewAsMap(event), "id"))
+        .instanceId(getString(getNewAsMap(event), "instanceId"))
+        .locationId(getString(getNewAsMap(event), "effectiveLocationId"))
```
viacheslavkol, line 83: Would be good to extract `getNewAsMap(event)` into a variable

src/main/java/org/folio/search/service/reindex/jdbc/CallNumberRepository.java
```diff
@@ -0,0 +1,184 @@
...
+        statement.setString(1, getId(entity));
+        statement.setString(2, getCallNumber(entity));
+        statement.setString(3, getPrefix(entity));
+        statement.setString(4, getSuffix(entity));
+        statement.setString(5, getTypeId(entity));
+        statement.setString(6, getVolume(entity));
+        statement.setString(7, getEnumeration(entity));
+        statement.setString(8, getChronology(entity));
+        statement.setString(9, getCopyNumber(entity));
+      });
```
viacheslavkol, line 126: Maybe one-by-one processing on batch update failure, same as for other resources?

src/main/java/org/folio/search/service/converter/preprocessor/extractor/impl/CallNumberResourceExtractor.java
```diff
@@ -0,0 +1,125 @@
...
+  @Override
+  public boolean hasChildResourceChanges(ResourceEvent event) {
+    if (!featureConfigService.isEnabled(TenantConfiguredFeature.BROWSE_CALL_NUMBERS)) {
+      return false;
+    }
+    var oldAsMap = getOldAsMap(event);
+    var newAsMap = getNewAsMap(event);
+    var oldCallNumber = constructEntity(oldAsMap);
+    var newCallNumber = constructEntity(newAsMap);
+    return !oldCallNumber.equals(newCallNumber);
```
mukhiddin-yusuf, line 67: perhaps it's better to make ```constructEntity``` return empty map instead of ```null```

src/main/java/org/folio/search/service/reindex/jdbc/CallNumberRepository.java
```diff
@@ -0,0 +1,184 @@
...
+    return getString(callNumberComponents, "callNumberTypeId");
+  }
+
+  private String getSuffix(Map<String, Object> callNumberComponents) {
+    return getString(callNumberComponents, "callNumberSuffix");
+  }
+
+  private String getPrefix(Map<String, Object> callNumberComponents) {
+    return getString(callNumberComponents, "callNumberPrefix");
+  }
```
mukhiddin-yusuf, line 183: extracting specific call number entity field from a map is also needed in other places I guess, like in CallNumberResourceExtractor, so maybe makes sense to unify them into some place and make it accessible.

psmagin: There is different maps in repository and in extractor, so methods will be different. I think we should refactor all this extractors and repositories to work with pojos and not maps.

mukhiddin-yusuf (approved):

viacheslavkol (approved):

PR #715, opened 2024-12-15
fix(change-ownership) fix "Invalid reference" appears after updating …

SvitlanaKovalova1:
### Purpose
[MSEARCH-915](https://folio-org.atlassian.net/browse/MSEARCH-915) The "Invalid reference" appears in the Location section for Holding after updating ownership.
When changing ownership the several events are generated:
-create holding/item for new tenant
-delete holding/item for old tenant.
Need to avoid deletion of all records with the same id for old and new tenant in the holding/item tables.

### Approach
- add tenant id to holding/item entity delete sql query.

### Changes Checklist

### Screenshots (if applicable)
image

### DIFF
--- a/src/main/java/org/folio/search/service/reindex/jdbc/MergeInstanceRepository.java
+++ b/src/main/java/org/folio/search/service/reindex/jdbc/MergeInstanceRepository.java
@@ -1,7 +1,6 @@
 package org.folio.search.service.reindex.jdbc;
 
 import static org.folio.search.utils.JdbcUtils.getFullTableName;
-import static org.folio.search.utils.JdbcUtils.getParamPlaceholderForUuid;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/search/service/reindex/jdbc/MergeInstanceRepository.java
```diff
@@ -81,4 +87,12 @@ public void updateBoundWith(String tenantId, String id, boolean bound) {
     var sql = UPDATE_BOUND_WITH_SQL.formatted(fullTableName);
     jdbcTemplate.update(sql, bound /*? "true" : "false"*/, id);
   }
+
+  @Override
+  public void deleteEntities(List<String> ids, String tenantId) {
```
mukhiddin-yusuf, line 92: is this ```deleteEntities``` overrides the one from ```MergeRangeRepository```? maybe it's better to have two separate delete methods, one with tenantId and one without?

SvitlanaKovalova1: updated

viacheslavkol (commented):
Would be good to have it covered with tests

viacheslavkol (approved):

mukhiddin-yusuf (approved):

PR #733, opened 2025-01-15
feat(cn-browse): support aliases for callNumberTypeId filters

psmagin:
### Purpose
Support call-number browse option names as aliases for callNumberTypeId filters for call-number documents.

### Approach
Add search term processor for call-number type id field.
Modify query construction for exact match term queries to support array of values.

### Changes Checklist

### Related Issues
https://folio-org.atlassian.net/browse/MSEARCH-942

### DIFF
--- a/src/main/java/org/folio/search/cql/searchterm/CallNumberTypeIdSearchTermProcessor.java
+++ b/src/main/java/org/folio/search/cql/searchterm/CallNumberTypeIdSearchTermProcessor.java
@@ -1,0 +1,25 @@
+package org.folio.search.cql.searchterm;
+
+import lombok.RequiredArgsConstructor;
+import org.folio.search.domain.dto.BrowseOptionType;
+import org.folio.search.domain.dto.BrowseType;
+import org.folio.search.service.consortium.BrowseConfigServiceDecorator;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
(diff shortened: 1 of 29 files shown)


src/main/java/org/folio/search/cql/searchterm/CallNumberTypeIdSearchTermProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+  private final BrowseConfigServiceDecorator configService;
+
+  @Override
+  public Object getSearchTerm(String inputTerm) {
+    try {
+      var browseOptionType = BrowseOptionType.fromValue(inputTerm);
+      var browseConfig = configService.getConfig(BrowseType.CALL_NUMBER, browseOptionType);
+      return browseConfig.getTypeIds().stream().map(Object::toString).toArray(String[]::new);
+    } catch (IllegalArgumentException e) {
+      return inputTerm;
```
mukhiddin-yusuf, line 22: don't we need to log some message here?

mukhiddin-yusuf (approved):

src/main/java/org/folio/search/cql/searchterm/CallNumberTypeIdSearchTermProcessor.java
psmagin: It's expected flow. If the input is not parsed to the enum then we proceed with input without any changes.
So, I believe no log msg needed

PR #771, opened 2025-03-11
fix(cn-browse): fix incorrect count for same call-number in one instance

psmagin:
### Purpose
fix incorrect count for same call-number in one instance 

### Changes Checklist

### Related Issues
[MSEARCH-992](https://folio-org.atlassian.net/browse/MSEARCH-992)

### DIFF
--- a/src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
+++ b/src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
@@ -2,7 +2,7 @@ package org.folio.search.service.browse;
 
 import static org.folio.search.utils.SearchUtils.CALL_NUMBER_TYPE_ID_FIELD;
 
-import java.util.List;
+import java.util.Collection;
 import java.util.Set;
 import java.util.function.Function;
 import lombok.extern.log4j.Log4j2;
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/search/service/browse/CallNumberBrowseService.java
```diff
@@ -55,33 +55,35 @@ protected BrowseResult<CallNumberBrowseItem> mapToBrowseResult(BrowseContext ctx
...
+                                  Function<CallNumberResource, Set<InstanceSubResource>> func) {
     return consortiumSearchHelper.filterSubResourcesForConsortium(ctx, resource, func)
       .stream()
       .map(InstanceSubResource::getInstanceId)
-      .map(List::size)
-      .reduce(0, Integer::sum);
+      .flatMap(Collection::stream)
+      .distinct()
+      .mapToInt(id -> 1)
+      .sum();
```
SvitlanaKovalova1, line 87: could we use `count()` here instead of `.mapToInt(id -> 1).sum()` ?

psmagin: `count()` return long value that underhood does 
`.mapToLong(e -> 1).sum();`
As we need to return int value, I decided to use `sum()`

SvitlanaKovalova1 (approved):

viacheslavkol (approved):

PR #788, opened 2025-04-25
Add flow diagrams for reindex process

okolawole-ebsco:
### Purpose
Add flow diagrams to reindex process to make onboarding easier.


psmagin (approved):
I think it is better to move `diagrams` folder to the existed `doc` folder.

okolawole-ebsco:
> I think it is better to move `diagrams` folder to the existed `doc` folder.

@psmagin done

viacheslavkol (approved):

PR #792, opened 2025-05-12
MSEARCH-1029: Implemented logic to truncate notes more than 32 000 characters long

vgema:
### Purpose
Implemented logic to truncate field values exceeding the allowed character limit to avoid indexing exceptions due to search engine constraints.

## Approach
Add String byte limit handling and custom serializer

### Changes Checklist

### DIFF
--- a/src/main/java/org/folio/search/service/setter/AbstractAllValuesProcessor.java
+++ b/src/main/java/org/folio/search/service/setter/AbstractAllValuesProcessor.java
@@ -77,7 +77,7 @@ public abstract class AbstractAllValuesProcessor implements FieldProcessor<Map<S
 
   private void collectFieldValuesFromEventBody(String path, MultilangValue ctx, Object v, Predicate<String> filter) {
     if (v instanceof String string) {
-      ctx.addValue(StringUtils.strip(string),
+      ctx.addValue(StringUtils.abbreviate(StringUtils.strip(string), MAX_FIELD_VALUE_LENGTH),
         searchFieldProvider.isFullTextField(ResourceType.INSTANCE, path));
     }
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/search/service/setter/AbstractAllValuesProcessor.java
```diff
@@ -77,7 +77,7 @@ private void collectFieldValuesFromEventBody(String path, MultilangValue ctx,
 
   private void collectFieldValuesFromEventBody(String path, MultilangValue ctx, Object v, Predicate<String> filter) {
     if (v instanceof String string) {
-      ctx.addValue(StringUtils.strip(string),
+      ctx.addValue(StringUtils.abbreviate(StringUtils.strip(string), MAX_FIELD_VALUE_LENGTH),
```
psmagin, line 80: I'm not sure that we need to use `abbreviate` method in this case. As a result of the method we will have a string with format "string value...", while three dots don't provide any improvement and will be not used while search (search engine will remove them and will not use them in indexing). I would suggest just cut string to the needed length.

psmagin (changes requested):
Please add unit tests and update NEWS file

psmagin (approved):

src/main/java/org/folio/search/service/setter/FieldProcessor.java
```diff
@@ -8,6 +8,8 @@
  */
 public interface FieldProcessor<T, R> {
 
+  int MAX_FIELD_VALUE_LENGTH = 32000;
+
```
SvitlanaKovalova1, line 12: I think it should be reduced to ~30,000 characters because when reindexing the note/administrativeNotes field, which originally contained 32,000 characters, we get the following error: 
image
https://folio-org.atlassian.net/browse/MSEARCH-1027

SvitlanaKovalova1 (approved):

viacheslavkol (approved):

PR #914, opened 2026-03-02
feat(reindex): add EXPORT-based merge ingestion for reindex using file-ready Kafka events and S3-backed record files 

psmagin:
### Purpose
Implement reindex EXPORT flow support and complete related refactoring/documentation updates for mod-search reindex processing.

This change is needed to support file-based merge ingestion (`inventory.reindex.file-ready`) in addition to publish-based ingestion, and to document/configure the new runtime knobs.

### Approach
- Added EXPORT-mode processing path for merge ingestion:
  - Consume `ReindexFileReadyEvent`
  - Read exported records from S3-compatible storage
  - Persist records to intermediate tables in batches
- Added `trace_id` support in merge ranges (`merge_range.trace_id`) and wired it into export request payloads.
- Introduced remote storage configuration/bean binding gated by `folio.reindex.reindex-type=EXPORT`.
- Added Kafka listener/container config for file-ready events and refactored reindex Kafka container-factory duplication.
- Refactored orchestration code to reduce duplication in merge success/failure handling and batch save logic.
- Updated diagrams and feature docs for full/upload flow, including PUBLISH vs EXPORT branch.
- Updated module descriptor + README environment variable documentation.

### Changes Checklist
- [ ] **API Changes**: No new public REST paths; existing reindex APIs unchanged.
- [x] **Database Schema Changes**: Added `trace_id` column to `merge_range` via Liquibase (`changes/v6.0/add-trace-id-range-column.xml`).
- [ ] **Interface Version Changes**: None.
- [x] **Interface Dependencies**: Added `org.folio:folio-s3-client` dependency.
- [ ] **Permissions**: No permission changes.
- [x] **Logging**: Added/updated logs for new reindex event paths and storage initialization.
- [x] **Unit Testing**: Updated/ran reindex orchestration unit tests.
- [ ] **Integration Testing**: Existing integration tests updated in codebase; full suite not run in this update session.
- [ ] **Manual Testing**: Not performed in this update session.
- [ ] **NEWS**: Not updated in this PR.

### Related Issues
[MSEARCH-1175](https://folio-org.atlassian.net/browse/MSEARCH-1175)

### Learning and Resources (if applicable)
- FOLIO RFC-0003 Breaking Changes guidance was used to assess impact classification.

### Screenshots (if applicable)
N/A


psmagin (commented):
## Code Review — MSEARCH-1175 EXPORT-based merge ingestion

The overall structure is solid — the Kafka factory refactoring is clean, the enum extraction is the right call, and the helper methods reduce duplication. There are a few issues that should be resolved before merging, detailed in inline comments below (and also in the summary at the end of this review body).

---

### Summary

| Priority | Count |
|----------|-------|
| ⚠️ High | 3 |
| 🟡 Medium | 3 |
| 🟢 Low | 2 |
| 📝 Probable FOLIO Breaking Changes | 1 |

**High** items to fix before merge:
1. `folio-s3-client:3.0.0-SNAPSHOT` — pin to a stable release
2. Unescaped `.` in `inventory\.reindex.file-ready` Kafka topic regex — matches unintended topics
3. Java-class default `reindexType = EXPORT` contradicts production YAML default of `PUBLISH`

**Medium** items:
- `folioS3Client` has no null guard when called in EXPORT path — NPE risk if misconfigured
- `event.getBucket()` is silently ignored; S3 client always uses its configured bucket regardless of the event field
- `handleMergeSuccess` reverses the original status-update order (counter before DB row update) — potential race with `isMergeCompleted()` check

**FOLIO Breaking Change (RFC-0003):** New environment variables and Kafka topic constitute an operational change — please bump the module major version and notify operators.

pom.xml
```diff
@@ -38,6 +38,7 @@
 
     <folio-spring-support.version>10.0.0</folio-spring-support.version>
     <folio-service-tools.version>6.0.0-SNAPSHOT</folio-service-tools.version>
+    <folio-s3-client.version>3.0.0-SNAPSHOT</folio-s3-client.version>
```
psmagin: ⚠️ **High — SNAPSHOT dependency in production code**

`folio-s3-client:3.0.0-SNAPSHOT` is a mutable artifact. Two builds on the same version string can resolve to different JARs, making releases non-reproducible. Please pin to a stable GA release before merging.

```xml
<folio-s3-client.version>3.0.0</folio-s3-client.version>
```

src/main/resources/application.yml
```diff
@@ -167,6 +168,10 @@ folio:
         concurrency: ${KAFKA_REINDEX_RECORDS_CONCURRENCY:2}
         topic-pattern: (${folio.environment}\.)(.*\.)inventory\.reindex-records
         group-id: ${folio.environment}-mod-search-reindex-records-group
+      reindex-file-ready:
+        concurrency: ${KAFKA_REINDEX_FILE_READY_CONCURRENCY:2}
+        topic-pattern: (${folio.environment}\.)(.*\.)inventory\.reindex.file-ready
```
psmagin: ⚠️ **High — Unescaped `.` in Kafka topic regex matches any character**

`inventory\.reindex.file-ready` — the `.` between `reindex` and `file-ready` is unescaped. In a Java regex `.` matches **any character**, so this consumer would also subscribe to topics like `inventory.reindexXfile-ready`. Every other pattern in this file correctly escapes its dots.

```yaml
# fix
topic-pattern: (${folio.environment}\.)(.*.\.)inventory\.reindex\.file-ready
```

src/main/java/org/folio/search/configuration/properties/ReindexConfigurationProperties.java
```diff
@@ -43,6 +43,13 @@ public class ReindexConfigurationProperties {
 
   private int mergeRangePublisherRetryAttempts = 5;
 
+  /**
+   * Defines a type of reindex process. Default is EXPORT.
+   * EXPORT - will use exported S3 files.
+   * PUBLISH - will use published Kafka messages.
+   */
+  private ReindexType reindexType = ReindexType.EXPORT;
```
psmagin: ⚠️ **High — Java-class default contradicts production YAML default**

The field is initialised to `ReindexType.EXPORT` here, but `application.yml` (production profile) defaults `REINDEX_TYPE` to `PUBLISH`. Any code path that does not load the YAML — including unit tests that construct this class directly — will silently behave as EXPORT.

```java
// fix
private ReindexType reindexType = ReindexType.PUBLISH;
```

Also, the Javadoc above says _"Default is EXPORT"_ — please update it to avoid misleading readers.

src/main/java/org/folio/search/service/reindex/ReindexOrchestrationService.java
```diff
@@ -21,56 +31,29 @@
...
-      log.debug("getMemberTenantIdForRangeProcessing:: Setting member tenant context: {}", memberTenantId);
-      ReindexContext.setMemberTenantId(memberTenantId);
-      return memberTenantId;
-    }
-
-    return null;
+  @Autowired(required = false)
+  public void setFolioS3Client(FolioS3Client folioS3Client) {
+    this.folioS3Client = folioS3Client;
   }
```
psmagin, line 44: 🟡 **Medium — No null guard + setter injection breaks immutability**

`folioS3Client` is injected via `@Autowired(required = false)`, so it is `null` in non-EXPORT deployments. If `process(ReindexFileReadyEvent)` is ever called when the client is `null` (misconfiguration, stale Kafka offset, etc.), `readAndSave` throws a bare `NullPointerException` with no diagnostic context.

Add a fast-fail guard at the top of `process(ReindexFileReadyEvent)`:

```java
if (folioS3Client == null) {
    throw new IllegalStateException(
        "FolioS3Client not configured; set folio.reindex.reindex-type=EXPORT");
}
```

Alternatively, accept `Optional<FolioS3Client>` via constructor injection to avoid the mutable-field pattern entirely.

src/main/java/org/folio/search/service/reindex/ReindexOrchestrationService.java
```diff
@@ -21,56 +31,29 @@
 @RequiredArgsConstructor
 public class ReindexOrchestrationService {
 
+  private static final int BATCH_SIZE = 500;
+
```
psmagin, line 19: 🟡 **Medium — Hardcoded `BATCH_SIZE` is not operator-tunable**

500 is a reasonable default, but for very large files or memory-constrained environments operators may need to tune this. Exposing it via `ReindexConfigurationProperties` would be consistent with how `mergeRangeBatchSize`, `uploadRangeSize`, etc. are already configured.

src/main/java/org/folio/search/utils/KafkaConstants.java
```diff
@@ -14,4 +14,5 @@ public class KafkaConstants {
   public static final String INSTANCE_SHARING_COMPLETE_LISTENER_ID = "mod-search-instance-sharing-complete-listener";
   public static final String REINDEX_RANGE_INDEX_LISTENER_ID = "mod-search-reindex-index-listener";
   public static final String REINDEX_RECORDS_LISTENER_ID = "mod-search-reindex-records-listener";
+  public static final String REINDEX_RECORDS_FILE_READY_ID = "mod-search-reindex-reindex-file-ready-listener";
```
psmagin: 🟢 **Low (nitpick) — `reindex` is duplicated in the listener ID string**

`"mod-search-reindex-reindex-file-ready-listener"` has `reindex` twice. Compare with `REINDEX_RECORDS_LISTENER_ID = "mod-search-reindex-records-listener"` which does not repeat the prefix.

```java
public static final String REINDEX_RECORDS_FILE_READY_ID = "mod-search-reindex-file-ready-listener";
```

src/main/java/org/folio/search/model/event/ReindexFileReadyEvent.java
```diff
@@ -0,0 +1,15 @@
+package org.folio.search.model.event;
+
+import lombok.Data;
+
+@Data
+public class ReindexFileReadyEvent {
+
+  private final String tenantId;
```
psmagin, line 8: 🟢 **Low — Inconsistent tenant field name + unused `bucket` field**

1. `ReindexFileReadyEvent` uses `tenantId` while `ReindexRecordsEvent` uses `tenant`. The `toReindexRecordsEvent` converter bridges this manually. Aligning on one name across both event types would prevent future mapping mistakes.

2. `event.getBucket()` is never used in `readAndSave` — the S3 client always reads from its configured bucket. Either use the event's bucket for the read call, or document why it is intentionally ignored (otherwise it looks like a bug if the event could reference a different bucket than what the module is configured for).

src/main/resources/changelog/changes/v6.0/add-trace-id-range-column.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.27.xsd">
+
+  <changeSet id="MSEARCH-1175@@alter:merge_range:add-column:trace_id" author="pavlo_smahin">
```
psmagin, line 8: 📝 **FOLIO Breaking Change — operational change (RFC-0003)**

This PR introduces new environment variables (`S3_REINDEX_URL`, `S3_REINDEX_REGION`, `S3_REINDEX_BUCKET`, `S3_REINDEX_ACCESS_KEY_ID`, `S3_REINDEX_SECRET_ACCESS_KEY`, `S3_REINDEX_IS_AWS`, `REINDEX_TYPE`, `KAFKA_REINDEX_FILE_READY_CONCURRENCY`) and a new Kafka topic dependency (`inventory.reindex.file-ready`). Per RFC-0003, **operational changes are breaking changes** that require a major implementation version bump. Please ensure the version is bumped and that release notes clearly document the new runtime requirements for operators.

src/main/java/org/folio/search/service/reindex/ReindexOrchestrationService.java
```diff
@@ -187,6 +242,37 @@ private ReindexException handleReindexUploadFailure(ReindexRangeIndexEvent event
...
+  private void handleMergeFailure(ReindexEntityType entityType, String rangeId, Exception ex) {
+    log.error(new FormattedMessage("process:: ReindexRecordsEvent indexing error [rangeId: {}, error: {}]",
+      rangeId, ex.getMessage()), ex);
+    reindexStatusService.updateReindexMergeFailed(entityType);
+    mergeRangeService.updateStatus(entityType, rangeId, ReindexRangeStatus.FAIL, ex.getMessage());
+  }
+
   private void handleReindexMergeFailure(ReindexRecordsEvent event, String errorMessage) {
     log.warn("handleReindexMergeFailure:: ReindexRecordsEvent indexing error [rangeId: {}, error: {}]",
       event.getRangeId(), errorMessage);
```
psmagin: 🟡 **Medium — `handleMergeSuccess` reverses the original status-update order**

The original `persistEntities` called `updateStatus(SUCCESS)` **then** `addProcessedMergeRanges`. The new helper reverses this: the processed-ranges counter is incremented first, then the DB row is marked SUCCESS. If `addProcessedMergeRanges` pushes the counter over the completion threshold and triggers `startUploadOnMergeCompletion()` before the range row is marked `SUCCESS`, a concurrent DB reader would see a range still `IN_PROGRESS` while the upload phase has already started.

Please confirm this reordering is intentional and safe, or restore the original order.

pom.xml
psmagin: fixed

src/main/resources/application.yml
psmagin: fixed

src/main/java/org/folio/search/configuration/properties/ReindexConfigurationProperties.java
psmagin: fixed

src/main/java/org/folio/search/service/reindex/ReindexOrchestrationService.java
psmagin: fixed

src/main/java/org/folio/search/utils/KafkaConstants.java
psmagin: fixed

src/main/resources/changelog/changes/v6.0/add-trace-id-range-column.xml
psmagin: Not a breaking change

src/main/java/org/folio/search/service/reindex/ReindexOrchestrationService.java
psmagin: fixed

src/main/java/org/folio/search/service/reindex/ReindexOrchestrationService.java
psmagin: Fixed by creating separate beans for each type

src/main/java/org/folio/search/model/event/ReindexFileReadyEvent.java
psmagin: Keep it as is

SvitlanaKovalova1 (approved):

PR #927, opened 2026-03-31
feat(instance-sharing): update the call number last_updated_date when handling the CONSORTIUM_INSTANCE_SHARING_COMPLETE event.

SvitlanaKovalova1:
### Purpose
Currently we story only one copy of instance in postgres and instance sharing on inventory storage produces CREATE event on central and DELETE event on member. During background processing there’s no definitive way to determine if instance is shared so call numbers are updated for every new instance on central consortium tenant. This could bring unnecessary load during big DI jobs.

### Approach
- Remove call number update from `InstanceChildrenResourceService.persistChildren`
- Add new Kafka listener for `CONSORTIUM_INSTANCE_SHARING_COMPLETE` event to update `last_updated_date` for call numbers when the sharing event occurs. 

### Changes Checklist

### Related Issues
[MSEARCH-1186](https://folio-org.atlassian.net/browse/MSEARCH-1186)


src/main/java/org/folio/search/integration/message/KafkaMessageListener.java
```diff
@@ -207,4 +252,15 @@ private static void logFailedEvent(IndexInstanceEvent event, Exception e) {
       event.tenant(), event.instanceId()
     ), e);
   }
+
+  private static void logFailedEvent(InstanceSharingCompleteEvent event, Exception e) {
+    if (event == null) {
+      log.warn("Failed to update tenantId for instance sharing complete event [event: null]", e);
```
psmagin: Not `tenantId` but `lastUpdatedDate`

psmagin (changes requested):

src/main/java/org/folio/search/configuration/kafka/InstanceSharingCompleteEventKafkaConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+
+  private final KafkaProperties kafkaProperties;
+
+  @Bean
+  public ConcurrentKafkaListenerContainerFactory<String, InstanceSharingCompleteEvent>
+    instanceSharingCompletedListenerContainerFactory() {
+
+    var factory = new ConcurrentKafkaListenerContainerFactory<String, InstanceSharingCompleteEvent>();
+    var deserializer = new JacksonJsonDeserializer<>(InstanceSharingCompleteEvent.class, false);
+    var overrideProperties = Map.<String, Object>of(MAX_POLL_RECORDS_CONFIG, 10);
```
viacheslavkol: Do we really need to poll in batches for this events? If so - I thing we need to add factory.setBatchListener(true);

psmagin (approved):

viacheslavkol (approved):

